PR #5, opened 2020-01-23
chore: add engines to the package.json

### LINKED ISSUE #4: Missing "engine" field in package.json
Worker threads are only available after Node.sj 10.x (out of memory), you probably want to add an "engine" field to "package.json" to issue a warning if someone tries to install it with an older version of Node.js

https://docs.npmjs.com/files/package.json#engines

ematipico:
Closes #4

### DIFF
--- a/package.json
+++ b/package.json
@@ -48,6 +48,6 @@
     "worker-threads-pool": "^2.0.0"
   },
   "engines": {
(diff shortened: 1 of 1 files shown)


pioardi:
Thanks for your help @vdeturckheim  @ematipico

PR #64, opened 2020-11-10
Rewrite to TypeScript

### LINKED ISSUE #62: Rewrite to TypeScript
I would like to rewrite the code to TypeScript and modify the build process to emit the lib folder on `build`
- With TypeScript we can identify potential problems and improve the code readability
- The TypeScript declarations will no longer differ from the code in the future, but will be generated natively

I'm interested to create a PR for that

Shinigami92:
closes #62


src/workers.ts
```diff
@@ -0,0 +1,108 @@
...
+        this.parent = value.parent
+      } else if (value.kill) {
+        // here is time to kill this thread, just clearing the interval
+        clearInterval(this.interval)
+        this.emitDestroy()
+      }
+    })
+  }
+
+  protected _checkAlive (): void {
```
Shinigami92, line 74: Return type in `index.d.ts` was boolean :thinking:  
Maybe just outdated

Shinigami92:
The devDependency `standard` currently only checks `tests` folder but not `src` folder  
We could try [`ts-standard`](https://github.com/standard/ts-standard)  
Or we could try [configure TypeScript for standard](https://standardjs.com/#typescript), but I think we need `eslint` for that? :thinking:  
Or we just relay on [`prettier`](https://prettier.io/docs/en/options.html)

Shinigami92:
Seems that the coverage step/job isn't working anymore
But this doesn't have anything to do with my changes

src/dynamic.ts
```diff
@@ -0,0 +1,71 @@
...
+          worker.postMessage({ kill: 1 })
+          worker.terminate()
+        }
+      })
+      return worker
+    }
+  }
+}
+
+module.exports = DynamicThreadPool
```
Shinigami92, line 71: These `module.exports` necessary to not break compatibility for now
Later we should remove this together with `"esModuleInterop": true` in `tsconfig.json` and introduce a breaking change release

Shinigami92:
@pioardi please have a look over the code and ask me anything you need to know
Then I will try my best to explain

src/dynamic.ts
jerome-benoit: Are you sure about that? 
The transpilation of export default class should build javascript module export compatible on its own given the right options.

src/workers.ts
jerome-benoit: I would remove the underscore on all private/protected methods/attributes since TS handle the scope.

src/dynamic.ts
```diff
@@ -0,0 +1,75 @@
+import { EventEmitter } from 'events'
+import { FixedThreadPoolOptions, WorkerWithMessageChannel } from './fixed'
+import FixedThreadPool from './fixed'
+
+class MyEmitter extends EventEmitter {}
+
+export type DynamicThreadPoolOptions = FixedThreadPoolOptions
```
jerome-benoit, line 7: Maybe putting type definitions is their own directory?

package.json
```diff
@@ -2,11 +2,10 @@
   "name": "poolifier",
   "version": "1.2.1",
   "description": "Library on top of node js worker threads that implement various worker pools type",
-  "main": "index.js",
-  "types": "index.d.ts",
+  "main": "lib/index.js",
   "scripts": {
-    "build": "npm install",
-    "test": "standard && nyc mocha --exit --timeout 20000 tests/*test.js ",
+    "build": "npm install && rimraf lib && tsc",
```
jerome-benoit, line 7: Integrating grunt for the build might be not difficult and will simplify scripts.

Shinigami92: We could do this (e.g. using [rollup](https://github.com/rollup/rollup)), but I think it's not really needed  
I personally prefer to stick to native tools if it's that easy like currently

src/dynamic.ts
Shinigami92: I don't think that's necessary. The types fit to the classes. And just create new files just for such few lines doesn't resolve any problems IMO

src/dynamic.ts
Shinigami92: Yes I'm sure and tried it locally  
If not using this (_legacy_) syntax, it is a breaking change
IMO we should make clean PRs/commits instead of just shift everything at once. In this way it's more clear what happend and it's more incrementally

src/workers.ts
Shinigami92: That's a good idea, but as I suggest in the commit with the `module.exports` I would prefer non breaking changes for this PR

pioardi:
@Shinigami92  @jerome-benoit  thanks a lot for all your work on that.
For the coverage step on the CI, it does not work when the project is forked the important thing is that all tests run well.

I will try to take a look more deeply as soon as I can, once that  this PR will be merged I will also give you more grants on this project :)

src/dynamic.ts
jerome-benoit: It makes the code cleaner by clearly separating types usage from types definition. And It was somehow the case before the rewrite to TS anyway. It also ease the types separation to @types package from the core features later on. Javascript only users can then have a smaller storage footprint.

src/workers.ts
jerome-benoit: It's not a breaking change, it only impacts private API ;)
That's not that important but for the sake of code cleanliness, I still consider it as an easy and valuable change to do at some point.

package.json
jerome-benoit: rollup, grunt, etc. and similar lightweight bundling tools (not like webpack ;)) have more advantages than drawbacks: build time, build size, build type, hooks handling, ...
Postponing its integration later on is possible but since it will be done at some point ...

jerome-benoit:
> The devDependency `standard` currently only checks `tests` folder but not `src` folder
> We could try [`ts-standard`](https://github.com/standard/ts-standard)
> Or we could try [configure TypeScript for standard](https://standardjs.com/#typescript), but I think we need `eslint` for that? 🤔
> Or we just relay on [`prettier`](https://prettier.io/docs/en/options.html)

I think switching to ts-standard is the right choice for now: one to one replacement, so easy and no extra works needed (hopefully).

src/dynamic.ts
Shinigami92: The TypeScript-specific part will end up in `d.ts` files  
If you haven't already, you should check out the branch and build it to see what happens
The `lib` folder is what actually lands in the npm package and the `src` folder is not delivered at all

So if I split the types into separate files, empty `js` files would be generated in addition to the `d.ts` files containing the types

If you really care about the bundle size, the only thing you should check out is the output of `npm publish --dry-run`

package.json
Shinigami92: That's a good idea, you should create a new issue for that task  
I would prefer rollup, because I heard gulp and grunt are dead/outdated tools and rollup is the way to go for library bundling (https://www.npmtrends.com/rollup-vs-grunt-vs-gulp)




I as a frontend developer just use `vue-cli` ^^ and for libraries like my prettier plugin I just use `tsc`
So that's why I don't really have any knowledge about lib-bundlers

jerome-benoit:
On top of that PR: 

+ ts-standard integration: https://github.com/Shinigami92/poolifier/pull/1/files 
+ (stacked over the previous PR) tslib integration: https://github.com/Shinigami92/poolifier/pull/2/files

Shinigami92:
I need to look why I leave the `typescript` dependency in the `package.json`  
https://github.com/pioardi/poolifier/pull/64/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R56

This doesn't look right to me, but I don't have the time for that in this moment

jerome-benoit:
> I need to look why I leave the `typescript` dependency in the `package.json`
> [#64 (files)](https://github.com/pioardi/poolifier/pull/64/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R56)
> 
> This doesn't look right to me, but I don't have the time for that in this moment

poolifier test just works without it.

package.json
jerome-benoit: In the meantime, could you please remove `npm install` from the build script if you do not plan to merge one of the PR I've made?
GitHub CI is then going to do it twice and build scripts always assume deps are installed.

src/fixed.ts
```diff
@@ -0,0 +1,144 @@
...
+    }
+  }
+
+  /**
+   * Execute the task specified into the constructor with the data parameter.
+   *
+   * @param data The input for the task specified.
+   * @returns Promise that is resolved when the task is done.
+   */
+  public async execute (data: Data): Promise<Response> {
```
Shinigami92, line 85: I didn't remove this `async` because I didn't want to change the code directly with this PR
But I think it can be removed because it's not necessary

This function just doesn't have an `await` in it

jerome-benoit: But it returns a Promise so it's a valid syntax and does not hurt at all the transpilation. 
And some coding style rules enforce the async keyword before methods or functions returning Promise.

src/dynamic.ts
jerome-benoit: There's still something that I do not understand: 
+ You have remove `module.exports.ThreadWorker = ThreadWorker` inside its class file;
+ You have kept `module.exports = DynamicThreadPool` and `module.exports = FixedThreadPool` inside their respective class files;
+ And there's `module.exports = { FixedThreadPool, DynamicThreadPool, ThreadWorker }` inside the index.ts file. 

I think the legacy module export syntax is only needed in the index.ts file and can be removed elsewhere.

pioardi:
@jerome-benoit @Shinigami92  just to inform you that i created a gitter room to chat and simplify all the discussions about poolifier.

I invited both of you, just in case you need please look at the readme where I pasted a link to the chat :)

package.json
```diff
@@ -2,16 +2,20 @@
...
-  "types": "index.d.ts",
+  "main": "lib/index.js",
   "scripts": {
-    "build": "npm install",
-    "test": "standard && nyc mocha --exit --timeout 20000 tests/*test.js ",
-    "debug-test": "mocha --inspect-brk --exit tests/*test.js ",
-    "demontest": "nodemon --exec \"npm test\"",
+    "build": "npm run build:clean && tsc",
+    "build:clean": "rimraf lib",
+    "test": "npm run build && nyc mocha --exit --timeout 20000 tests/*test.js ",
```
pioardi, line 9: The standard command is no more present into the test script, standard is a main component to ensure  the quality of this project

src/workers.ts
pioardi: I agree with @jerome-benoit , we don't really care about breaking changes for private methods, end users should not use private methods and that's why we prefixed them with  "_", one benefit of typescript is that we can declare private methods and I would proceed with that.

pioardi:
I did some improvements to the repo management.

 - I added both of you @Shinigami92  @jerome-benoit  as collaborators to this repo so that you can create a branch without a fork ( this will allow you to use also the coverage step).
 -I added restrictions to the master branch, you will need to go through a PR to merge the code on the master branch ( I am excluded from that since that I am the administrator of the project).
 - I removed the step for coverage when a PR is open, we could probably need to look into that better.
 - I will add now the check for passing the npm run test script to merge a PR

package.json
jerome-benoit: If you expect standard to run without error in test script, then ts-standard should also be run here now.

Shinigami92: Yeah, I'm currently working on the branch
Give me some time to respond to all of the new stuff from this night ^^

Shinigami92:
supersede by #72

package.json
pioardi: Ts-standard Is a good option if the corrente standard does not work proprrly for typescript :) It Is ok for me

PR #72, opened 2021-02-07
Rewrite to TypeScript

### LINKED ISSUE #62: Rewrite to TypeScript

Shinigami92:
closes #62


Shinigami92:
The `Verify scale thread up and down is working` test in `Dynamic thread pool test suite` is failing :warning:

Shinigami92:
Ups, while rebasing I think it missed this part
Now it's working again

Shinigami92:
@pioardi I would suggest that we merge this together in its current state and move on from there

We have to set the ts compiler to `strict` and can start changing the code
However, currently this PR is a "take the JS code and add TS types and TS constructor optimizations"

Also this PR is a non-breaking backwards compatible branch to the current package version `1.2.1`

New changes should and will be made, but they are no longer backward compatible. So we should create new branches for them so we can do better reviews and cleaner commits.

pioardi:
I see a lot of issues when running the npm run test script, in particular with the ts-standard task more than others.
I would suggest to have at least all the test script working properly before to merge on master.

package.json
```diff
@@ -2,16 +2,20 @@
...
-  "types": "index.d.ts",
+  "main": "lib/index.js",
   "scripts": {
-    "build": "npm install",
-    "test": "standard && nyc mocha --exit --timeout 20000 tests/*test.js ",
-    "debug-test": "mocha --inspect-brk --exit tests/*test.js ",
-    "demontest": "nodemon --exec \"npm test\"",
+    "build": "npm run build:clean && tsc",
+    "build:clean": "rimraf lib",
+    "test": "standard && ts-standard && npm run build && nyc mocha --exit --timeout 20000 tests/*test.js ",
```
pioardi, line 9: @jerome-benoit I guess we an just use ts-standard and remove the first standard command here ?

Shinigami92:
> Said that I agree to merge this as soon as we can, but at least we need to fix the test script issues before to merge

Okay, then I'll create a branch based on this PR and address #73 with that

package.json
```diff
@@ -2,16 +2,20 @@
...
+    "test:debug": "mocha --inspect-brk --exit tests/*test.js ",
+    "test:nodemon": "nodemon --exec \"npm test\"",
     "coverage": "nyc report --reporter=text-lcov | coveralls",
-    "standard-verbose": "npx standard --verbose",
-    "lint": "prettier-standard --format | standard --fix"
+    "import-sort": "npx import-sort-cli --write 'src/**/*.ts{,x}'",
+    "standard:verbose": "npx standard --verbose",
+    "lint:standard": "prettier-standard --format | standard --fix",
```
pioardi, line 15: I would like to have one  script called "lint" that do everything needed to pass the test script.
Based on the above comment we could need only ts-standard task to run during the test phase, and we can also remove the lint-and-format, lint:standard, lint:s-standard and just replace all of them with one script "lint" that should look like this : 

    "lint": "ts-standard --fix  && npm run import-sort"

pioardi (changes requested):
We are really close to merge this and move ahead, I will try as soon as I can to fix the npm run test script that we have currently and we could have a good version to merge on master

package.json
Shinigami92: Can we please do this in separate branches, otherwise it will be really chaotic here

package.json
Shinigami92: Can we please do this in separate branches, otherwise it will be really chaotic here

package.json
jerome-benoit: I tend to agree with @Shinigami92, ts-standard, even if based on standard seems to behave differently on javascript code. 
The goal of its introduction was to: 
+ keep a known working linter setup for all files except TS files;
+ add a similar linter setup to deal with TS files and standard misbehave with them.

package.json
```diff
@@ -46,12 +66,13 @@
...
+    "rimraf": "^3.0.2",
+    "standard": "^16.0.3",
+    "ts-standard": "^10.0.0",
+    "typescript": "^4.1.3"
   },
   "engines": {
     "node": ">=12.0.0"
   },
-  "dependencies": {
-    "typescript": "^4.1.3"
```
jerome-benoit, line 55: That means that the typescript runtime dep was wrongly there before the TS rewrite :)

package.json
jerome-benoit: And for debugging purpose of each linter, having separate scripts to run them is currently useful.

src/dynamic.ts
```diff
@@ -0,0 +1,78 @@
...
+          this.workers.splice(workerIndex, 1)
+          this.tasks.delete(worker)
+        }
+      })
+      return worker
+    }
+  }
+}
+
+module.exports = DynamicThreadPool
```
jerome-benoit, line 78: There's still something that I do not understand:

+ You have removed module.exports.ThreadWorker = ThreadWorker inside its class file;
+ You have kept module.exports = DynamicThreadPool and module.exports = FixedThreadPool inside their respective class files;
+ And there's module.exports = { FixedThreadPool, DynamicThreadPool, ThreadWorker } inside the index.ts file.

I think the legacy module export syntax is only needed in the index.ts file and can be removed elsewhere.

Shinigami92: I tried the library with a productive project before and after the rewrite to typescript
Without these `module.exports` for `DynamicThreadPool` and `FixedThreadPool` the user needs to use another import statement syntax as before
That's what I mean with a breaking API change
So I added these `module.exports` where necessary so the package exports the same _API_ as without the typescript rewrite
And yes we should remove these exports with #74 

After merging #77 it could be possible to release a last v1.3.0 or so. This way at least dependend projects using `1.x.x` benefit also from the typescript rewrite

Then we can build a `2.x.x` and if you want so we can pack any feature and optimizations into this as you want

jerome-benoit: I'm perfectly fine the module.exports usage for backward compatibility. 
But I just do not understand why they are needed in all these places. 

Isn't `module.exports = { FixedThreadPool, DynamicThreadPool, ThreadWorker } `replacing all the others module.exports? 
Have `module.exports = DynamicThreadPool` a different semantic than `module.exports = { DynamicThreadPool }`? 
I'm not very familiar with that legacy syntax and its semantic.

Shinigami92: https://stackoverflow.com/a/40295288/6897682

As I said, for v2 we will remove all these `module.exports` and just relay on `export / export default`

jerome-benoit: That reads, are you sure then that you can remove safely `module.exports.ThreadWorker = ThreadWorker` from its class file? 

Thanks.

Shinigami92: Yes, because it's not `module.exports = ThreadWorker` but `module.exports.ThreadWorker = ThreadWorker`

And this is the equivalent of `export class ThreadWorker ...`

src/workers.ts
```diff
@@ -1,18 +1,44 @@
...
+  protected lastTask: number
+  protected readonly interval: NodeJS.Timeout
+  protected parent: any
+
+  public constructor (
+    fn: (data: Data) => Response,
+    public readonly opts: ThreadWorkerOptions = {}
+  ) {
     super('worker-thread-pool:pioardi')
-    this.opts = opts || {}
```
pioardi, line 15: Why did we removed this line ? Any particular reason ?

Shinigami92: It was just moved to line 38

pioardi: Thanks @Shinigami92

pioardi:
I would prefer to fix the ts-standard lints before to move on, but let me know if for you is ok or if we want to merge and work on lints later on.

Shinigami92:
@pioardi 

> I would prefer to fix the ts-standard lints before to move on, but let me know if for you is ok or if we want to merge and work on lints later on.

I already have created this PR: https://github.com/pioardi/poolifier/pull/77

I'm currently in a meeting and after that I will resolve the package.json conflict here

I would prefer to merge this PR into master separately to https://github.com/pioardi/poolifier/pull/77 because this way, we have a more clean history and don't have too much changes between before TS rewrite and after TS rewrite  
It just make it hard to track changes between js->ts changes

---

And after #77 was merged, I wont stop contributing ^^
It's just the beginning :slightly_smiling_face:

pioardi:
Sounds good @Shinigami92 , I will let it go to the master even if we will not release a new version for now :)

pioardi:
@Shinigami92  @jerome-benoit  Thanks a lot for all your work on this, I am sure with your help this project can grow further.

I have one question for both, which are the real benefits from an end-user prespective of re-write all the files to typescript?
I mean, we already have typescript definitions in place and people can use this library with both js and ts already, and I also see a lot of libraries that do in this way, write in pure js and provide types to support ts.

Which is the real benefit of write the project fully in typescript ? I am asking that also because if we have any further benefit we should document them. 
For what I know write everyhing in ts could increase the performances, so I am thinking to measure the two versions :)

One of the pillars of this project is performance.
Other pillars are:
  -  Easy to use
  -  Secure and with few runtime dependencies
  -  In Quality ( easy to mantain )
  -  Performance ( we want this project to have the best performance over other alternatives )

pioardi (approved):
We are ready to go @Shinigami92  @jerome-benoit  thanks a lot for your interest on that.
We are going to merge on the master but for now we will not release a version, there are a couple of things that we want to fix before to do a release, I will try to get a backlog if needed

Shinigami92:
> I have one question for both, which are the real benefits from an end-user prespective of re-write all the files to typescript?

With native written TS your don't miss any types + you are definitely sure that you don't have null-check issues and so on
`.d.ts` can potentially serve wrong types (like see the `_checkAlive (): boolean` but it was `_checkAlive (): void`)

> In Quality ( easy to mantain )

With first-class/native TS support also the maintainability increases, because you can benefit from the `strict` flag

> Performance

TS should not affect performance directly, because theoretically it just adds types that are remove while transpiling  
If there are performance issues, there is always a way to also write correct JS code, TS is just JS with types. Non of these types are in the transpiled js files in lib folder

jerome-benoit:
> Which is the real benefit of write the project fully in typescript ? I am asking that also because if we have any further benefit we should document them.

The typing as long as you strictly forbid 'any' usage ;), the OOP with generics, useful syntax shortcuts, ... 

> For what I know write everyhing in ts could increase the performances, so I am thinking to measure the two versions :)

Well, the transpilater is smart but not smart enough to make crappy TS code that do not scale into clean JS code that scale :)

> One of the pillars of this project is performance.
> Other pillars are:
> 
> * Easy to use
> * Secure and with few runtime dependencies

No impact on it. TS only requires dev deps.

> * In Quality ( easy to maintain )

TS code is definitely easier to maintain because of OOP full support and less error prone due to type checking, logic checking. 

> * Performance ( we want this project to have the best performance over other alternatives )

No impact, the JS generated can even be more a bit more optimised.

pioardi:
Really good, one of the things that we can do is try to define an interface ( is already there, the execute method) that is implemented in all of our pools so that people can easily switch from a pool to another.

Strategy Design Pattern basically, that was not possible with js ( I did it with a trick but with ts can be improved )

jerome-benoit:
> Really good, one of the things that we can do is try to define an interface ( is already there, the execute method) that is implemented in all of our pools so that people can easily switch from a pool to another.
> 
> Strategy Design Pattern basically, that was not possible with js ( I did it with a trick but with ts can be improved )

Well, you can define an abstract class that define all the shared attributes, methods and methods signature that each pool type must implement or resuse.  
Not sure it worth more effort to stick to a strict strategy design pattern implementation. With the an abstract class, you can already factorise some code between pool implementation and enforce the pool implementation to follow an API.

Issue #75, opened 2021-02-07
Rename functions and methods to not use prefix underscore


Shinigami92:
followup of #62 

we now use typescript with member accessors, so we can get rid of these prefixes
theoretically this is a breaking change, but users wont be really affected, because they shouldn't use these methods anyway

Issue #74, opened 2021-02-07
Remove usage of module.exports


Shinigami92:
followup of #62 

this will introduce breaking changes to the API :warning: 
we need to add a "how to migrate" (or something like that) to the `CHANGELOG.md`

PR #77, opened 2021-02-07
Use strict compiler flag in tsconfig

### LINKED ISSUE #73: Use strict compiler flag in tsconfig
followup of #62 

this will also fix most of the ts-standard problems currently occur in #72

Shinigami92:
closes #73 

Attention :warning: this branch should be merged directly into master AFTER #72 was merged
I will switch the base to master after #72 was merged into master

### DIFF
--- a/src/dynamic.ts
+++ b/src/dynamic.ts
@@ -1,3 +1,5 @@
+/* eslint-disable @typescript-eslint/strict-boolean-expressions */
+
 import FixedThreadPool, {
   FixedThreadPoolOptions,
   WorkerWithMessageChannel
@@ -42,7 +44,7 @@ export default class DynamicThreadPool<
   }
 
   protected _chooseWorker (): WorkerWithMessageChannel {
-    let worker: WorkerWithMessageChannel
+    let worker: WorkerWithMessageChannel | undefined
     for (const entry of this.tasks) {
       if (entry[1] === 0) {
         worker = entry[0]
@@ -60,10 +62,11 @@ export default class DynamicThreadPool<
       }
       // all workers are busy create a new worker
       const worker = this._newWorker()
-      worker.port2.on('message', (message) => {
+      worker.port2?.on('message', (message: { kill?: number }) => {
         if (message.kill) {
           worker.postMessage({ kill: 1 })
-          worker.terminate()
+          // eslint-disable-next-line no-void
+          void worker.terminate()
           // clean workers from data structures
           const workerIndex = this.workers.indexOf(worker)
           this.workers.splice(workerIndex, 1)
--- a/src/fixed.ts
+++ b/src/fixed.ts
@@ -91,7 +91,12 @@ export default class FixedThreadPool<Data = any, Response = any> {
   public execute (data: Data): Promise<Response> {
     // configure worker to handle message with the specified task
     const worker = this._chooseWorker()
-    this.tasks.set(worker, (this.tasks.get(worker) ?? 0) + 1)
+    const previousWorkerIndex = this.tasks.get(worker)
+    if (previousWorkerIndex !== undefined) {
+      this.tasks.set(worker, previousWorkerIndex + 1)
+    } else {
+      throw Error('Worker could not be found in tasks map')
+    }
     const id = ++this._id
     const res = this._execute(worker, id)
     worker.postMessage({ data: data || _void, _id: id })
@@ -111,7 +116,12 @@ export default class FixedThreadPool<Data = any, Response = any> {
       }): void => {
         if (message._id === id) {
           worker.port2?.removeListener('message', listener)
-          this.tasks.set(worker, (this.tasks.get(worker) ?? 0) - 1)
+          const previousWorkerIndex = this.tasks.get(worker)
+          if (previousWorkerIndex !== undefined) {
+            this.tasks.set(worker, previousWorkerIndex + 1)
+          } else {
+            throw Error('Worker could not be found in tasks map')
+          }
           if (message.error) reject(message.error)
           else resolve(message.data)
         }
--- a/src/fixed.ts
+++ b/src/fixed.ts
@@ -1,6 +1,8 @@
+/* eslint-disable @typescript-eslint/strict-boolean-expressions */
+
 import { MessageChannel, SHARE_ENV, Worker, isMainThread } from 'worker_threads'
 
-function empty () {}
+function empty (): void {}
 const _void = {}
 
 export type Draft<T> = { -readonly [P in keyof T]?: T[P] }
@@ -41,10 +43,12 @@ export default class FixedThreadPool<Data = any, Response = any> {
   public nextWorker: number = 0
 
   // threadId as key and an integer value
+  /* eslint-disable @typescript-eslint/indent */
   public readonly tasks: Map<WorkerWithMessageChannel, number> = new Map<
     WorkerWithMessageChannel,
     number
   >()
+  /* eslint-enable @typescript-eslint/indent */
 
   protected _id: number = 0
 
@@ -61,6 +65,7 @@ export default class FixedThreadPool<Data = any, Response = any> {
     if (!isMainThread) {
       throw new Error('Cannot start a thread pool from a worker thread !!!')
     }
+    // TODO christopher 2021-02-07: Improve this check e.g. with a pattern or blank check
     if (!this.filePath) {
       throw new Error('Please specify a file with a worker implementation')
     }
@@ -82,30 +87,36 @@ export default class FixedThreadPool<Data = any, Response = any> {
    * @param data The input for the task specified.
    * @returns Promise that is resolved when the task is done.
    */
-  public async execute (data: Data): Promise<Response> {
+  // eslint-disable-next-line @typescript-eslint/promise-function-async
+  public execute (data: Data): Promise<Response> {
     // configure worker to handle message with the specified task
     const worker = this._chooseWorker()
-    this.tasks.set(worker, this.tasks.get(worker) + 1)
+    this.tasks.set(worker, (this.tasks.get(worker) ?? 0) + 1)
     const id = ++this._id
     const res = this._execute(worker, id)
     worker.postMessage({ data: data || _void, _id: id })
     return res
   }
 
+  // eslint-disable-next-line @typescript-eslint/promise-function-async
   protected _execute (
     worker: WorkerWithMessageChannel,
     id: number
   ): Promise<Response> {
     return new Promise((resolve, reject) => {
-      const listener = (message) => {
+      const listener = (message: {
+        _id: number
+        error?: string
+        data: Response
+      }): void => {
         if (message._id === id) {
-          worker.port2.removeListener('message', listener)
-          this.tasks.set(worker, this.tasks.get(worker) - 1)
+          worker.port2?.removeListener('message', listener)
+          this.tasks.set(worker, (this.tasks.get(worker) ?? 0) - 1)
           if (message.error) reject(message.error)
           else resolve(message.data)
         }
       }
-      worker.port2.on('message', listener)
+      worker.port2?.on('message', listener)
     })
   }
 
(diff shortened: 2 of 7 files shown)


src/dynamic.ts
```diff
@@ -42,15 +42,15 @@ export default class DynamicThreadPool<
...
+    let worker: WorkerWithMessageChannel | undefined
     for (const entry of this.tasks) {
       if (entry[1] === 0) {
         worker = entry[0]
         break
       }
     }
 
-    if (worker) {
+    if (worker != null) {
```
jerome-benoit, line 53: I would really prefer to have an helper method for all these null or undefined check. 

A file like utils.ts with: 

```
export function isNullOrUndefined<T>(object: T): boolean {
  return object == null;
}
```

And replacing such checks by:

`!isNullOrUndefined(worker)`
...


Or an Utils class containing only static methods as helpers.

jerome-benoit, line 53: I would really prefer to have an helper method for all these null or undefined check. 

A file like utils.ts with: 

```
export function isNullOrUndefined<T>(object: T): boolean {
  return object == null;
}
```

And replacing such checks by:

`!isNullOrUndefined(worker)`
...


Or an Utils class containing only static methods as helpers.

Shinigami92: I would prefer to disable this eslint rule and use `!object` checks
They are not even smaller and common used, but also they are `falsy` and not `null/undefined` checks
So e.g. for the `!this.filePath` it also checks for empty string

If @pioardi really want something like util functions like `isNullOrUndefined` we should use a library for that. e.g. [isNil](https://github.com/justinlettau/ts-util-is/blob/master/src/index.ts#L181)

But really... I think this is absolutely overkill

src/fixed.ts
```diff
@@ -61,6 +63,8 @@ export default class FixedThreadPool<Data = any, Response = any> {
     if (!isMainThread) {
       throw new Error('Cannot start a thread pool from a worker thread !!!')
     }
+    // TODO christopher 2021-02-07: Improve this check e.g. with a pattern or blank check
+    // eslint-disable-next-line @typescript-eslint/strict-boolean-expressions
     if (!this.filePath) {
```
jerome-benoit, line 68: If you use `typeof this.filePath === undefined`, that would fix the linter.

Shinigami92: And fail the test... yeah...
It's not the same

Please try 
```js
!'' // true
typeof '' === undefined // false
```

src/dynamic.ts
jerome-benoit: > I would prefer to disable this eslint rule and use `!object` checks
> They are not even smaller and common used, but also they are `falsy` and not `null/undefined` checks
> So e.g. for the `!this.filePath` it also checks for empty string

If ts-standard is using eslint + TS plugin behind the scene, there should be a way to disable some rule globally.

Shinigami92: But I would prefer to let decide @pioardi if we should do that, because it would be introducing another config file to the project

Shinigami92: Looks like a duplicate review, I will resolve this one

jerome-benoit: ts-standard documentation: 

🚫 Please change X rule

This project has no control over the rules implemented, as such this project cannot change any of the rules that have been configured. If you want to discuss the rules, please visit the rules configuration repo eslint-config-standard-with-typescript.

pioardi:
@jerome-benoit  I guess we are ready to finalize this PR.
Typescript stuff is now on master so you will need to take the laster changes from master

pioardi:
Please also take a look at the project section https://github.com/pioardi/poolifier/projects/1 
We coul realize a part of those improvements before to release a new version ( 2.x.x or a 1.x.x will depends on the backward compatibility )

Shinigami92:
:eyes: test are failing, I will have a look into that

Shinigami92:
Ok... tests working again, seems it just was a hiccup ^^'

Shinigami92:
@pioardi when you merge this right now, I will start creating PRs for #74 and #75

Shinigami92:
> We coul realize a part of those improvements before to release a new version ( 2.x.x or a 1.x.x will depends on the backward compatibility )

1eaa679e42a2a158b38e6fd615fd04b1bf6eb694 is the last commit that is **backward compatibility**
I would suggest that every commit after that will be for 2.x.x

src/dynamic.ts
pioardi: Oh I honestly I am not sure on this.
I am more familiar with the original js syntax like 

`if(!this.x)`

Now I understand that with typescript this could not be the best practice, so if we are concerned from that we can just write an util method.

No external libraries please for this tiny thing, this project runs without other 3rd part libraries and unless that introduce a library enable us for something great we should avoid them.

Shinigami92: Okay we should disable the lint rule for that, I will try...

pioardi:
Hei @Shinigami92  why that commit is not backward compatible ? We run our tests without changing so I see it should be backward compatible. 
Do you refer to the methods prefixed with "_" , I don't think anyone used that methods but yes we could consider that anyway and release a version 2.x.x.

We will decide before the release, I really want to add also some  features that are into the project section ( ScheduledThreadPool, ForkJoinPool )

src/fixed.ts
```diff
@@ -61,6 +65,8 @@ export default class FixedThreadPool<Data = any, Response = any> {
     if (!isMainThread) {
       throw new Error('Cannot start a thread pool from a worker thread !!!')
     }
+    // TODO christopher 2021-02-07: Improve this check e.g. with a pattern or blank check
+    // eslint-disable-next-line @typescript-eslint/strict-boolean-expressions
```
jerome-benoit: I think you can remove that rule disablement now that it's done at the file level.

Shinigami92:
I'm curious but it seems that when running `npm run lint:standard` the "standard.ignore" section in `package.json` doesn't take any effect

Shinigami92:
@pioardi Okay... sadly it's not possible to tell `prettier-standard` to ignore just `src/**/*.ts` files :confused:  
So usages of `import type` will be removed by using `prettier-standard` :shrug:  

Also I need to explicitly disable rules in files to ignore them. With standard/ts-standard it's not possible to configure rules via e.g. `.eslintrc.json` :confused:

pioardi:
So what is not so clear to me is how our code should looks like to avoid this

src/fixed.ts
```diff
@@ -82,30 +87,36 @@ export default class FixedThreadPool<Data = any, Response = any> {
    * @param data The input for the task specified.
    * @returns Promise that is resolved when the task is done.
    */
-  public async execute (data: Data): Promise<Response> {
+  // eslint-disable-next-line @typescript-eslint/promise-function-async
+  public execute (data: Data): Promise<Response> {
     // configure worker to handle message with the specified task
     const worker = this._chooseWorker()
-    this.tasks.set(worker, this.tasks.get(worker) + 1)
+    this.tasks.set(worker, (this.tasks.get(worker) ?? 0) + 1)
```
pioardi, line 94: we should fail if this.tasks.get(worker) does not return something.....That means that something was wrong...

src/fixed.ts
```diff
@@ -82,30 +87,36 @@ export default class FixedThreadPool<Data = any, Response = any> {
    * @param data The input for the task specified.
    * @returns Promise that is resolved when the task is done.
    */
-  public async execute (data: Data): Promise<Response> {
+  // eslint-disable-next-line @typescript-eslint/promise-function-async
+  public execute (data: Data): Promise<Response> {
```
pioardi, line 91: should async keywork be there ?

src/fixed.ts
```diff
@@ -82,30 +87,36 @@ export default class FixedThreadPool<Data = any, Response = any> {
...
+      const listener = (message: {
+        _id: number
+        error?: string
+        data: Response
+      }): void => {
         if (message._id === id) {
-          worker.port2.removeListener('message', listener)
-          this.tasks.set(worker, this.tasks.get(worker) - 1)
+          worker.port2?.removeListener('message', listener)
+          this.tasks.set(worker, (this.tasks.get(worker) ?? 0) - 1)
```
pioardi, line 114: same as above

Shinigami92:
> So what is not so clear to me is how our code should looks like to avoid this ?
> 
> `/* eslint-disable @typescript-eslint/strict-boolean-expressions */ `

Don't use standard, prettier-standard and ts-standard but just relay on `eslint` and `prettier(x)` itself and configure it to your needs  
Also you can profit then from first-class support of IDE-support without extensions (if I'm not wrong ^^, I have these two extensions installed: `dbaeumer.vscode-eslint` and `esbenp.prettier-vscode`)

pioardi (changes requested):

src/fixed.ts
Shinigami92: Not really, we return a **raw** promise here and don't have the `await` keyword in use in the body of this function

There is even a rule for that https://eslint.org/docs/rules/require-await

src/fixed.ts
Shinigami92: Okay, so we have two options here
1. Using the `I know it better operator` (`this.tasks.set(worker, this.tasks.get(worker)! + 1)`)
2. splitting it into lines and throw Error if it's broken
e.g.: (just pseudo-code)
```ts
const prev = this.tasks.get(worker)
if (prev)
  this.tasks.set(worker, prev + 1)
else throw Error()
```

Shinigami92:
> > So what is not so clear to me is how our code should looks like to avoid this ?
> > `/* eslint-disable @typescript-eslint/strict-boolean-expressions */ `
> 
> No falsy generic test anymore, you have to specifically test for `=== null && === '' && typeof === 'undefined'` etc.

Or
- `if (typeof fn !== 'function') throw Error`
- `if (typeof this.filePath !== 'string' || this.filePath === '') throw Error`

and so on

pioardi:
> No falsy generic test anymore, you have to specifically test for `=== null && === '' && typeof === 'undefined'` etc.

So I guess we will need that util method that you mentioned. 

But why ts-standard is requesting this check when we do an import....?  I am sorry I am not familiar with typescript I will need a good course on that :D

src/fixed.ts
pioardi: I would prefer the second much more clear for me and with a clear message on the error

Shinigami92:
> > No falsy generic test anymore, you have to specifically test for `=== null && === '' && typeof === 'undefined'` etc.
> 
> So I guess we will need that util method that you mentioned.

Or we could just write it inline in the code
... But I would suggest to setup a good lint+format workflow first
I will propose one after this PR is merged. Hope you can then see what's the benefit of using the more native stuff :slightly_smiling_face: 
 
> But why ts-standard is requesting this check when we do an import....? I am sorry I am not familiar with typescript I will need a good course on that :D

Because ts-standard is not up-to-date and behind it's possibilities
`import type` is a TS 3.8 feature https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-8.html

src/fixed.ts
```diff
@@ -82,30 +87,36 @@ export default class FixedThreadPool<Data = any, Response = any> {
...
   }
 
+  // eslint-disable-next-line @typescript-eslint/promise-function-async
   protected _execute (
     worker: WorkerWithMessageChannel,
     id: number
   ): Promise<Response> {
     return new Promise((resolve, reject) => {
-      const listener = (message) => {
+      const listener = (message: {
```
Shinigami92, line 107: I will extract an interface/type for the Message-object in a later PR

src/fixed.ts
pioardi: I would prefer to stay with async keyword since that was previously present

src/fixed.ts
Shinigami92: Could you please tell me what `prev` really is :sweat_smile:  
And if you could also provide me a good error message, that would be awesome

src/fixed.ts
Shinigami92: Oh... now it wants a `return await res`, but we don't want to await that here...
And this is coming from `ts-standard` and this is not really up-to-date :confused:

Shinigami92: So to explain what happens here:
we have `const res = this._execute(worker, id)` so res is a raw Promise
and this raw Promise is returned by the function
if we use async await, we could also write `const res = await this._execute(worker, id)`
But maybe we don't want that, because we want to decide the user of our tool when he wants to await the response on its own

src/fixed.ts
jerome-benoit: It's the worker index number.

src/fixed.ts
jerome-benoit: I vote for removing the async, since it makes life so much easier without any drawbacks.

pioardi: Just thinking if removing the async can break compatibility for end-users mmmmmm
I guess not right? Then we can just remove it

jerome-benoit (approved):

pioardi (approved):
This is ready to go, we will need to add some tests to cover more code branches or just decrease the target coverage percentage.
But we will handle that going on :)

PR #81, opened 2021-02-08
Ensure all commands are always run in lint-and-format

jerome-benoit:

### DIFF
--- a/examples/typescript/pool.ts
+++ b/examples/typescript/pool.ts
@@ -1,6 +1,7 @@
-import { join } from 'path'
 import { DynamicThreadPool, FixedThreadPool } from 'poolifier'
 import { MyData, MyResponse } from './worker'
+
+import { join } from 'path'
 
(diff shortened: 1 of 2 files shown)


examples/typescript/pool.ts
```diff
@@ -1,7 +1,8 @@
-import { join } from 'path'
 import { DynamicThreadPool, FixedThreadPool } from 'poolifier'
 import { MyData, MyResponse } from './worker'
 
+import { join } from 'path'
```
pioardi, line 4: Please just clean up spaces not needed

jerome-benoit: You mean blank lines? 

The default setup that sort imports put a line separator between pure node.js module import and the other modules. I have to look on how

pioardi: That's ok, we don't need to look into that and complicate the things :)

pioardi (approved):

PR #85, opened 2021-02-08
Use prettierx and eslint native

Shinigami92:
closes #76


pioardi:
@Shinigami92  I am not sure with this, I really like standard ( ts-standard I little less lol ) because we can check the code and see if is well formatted and meet all the best practices into our tests.
Is also used widely in a lot of js projects

Shinigami92:
And NOW this PR comes into your repo
Providing you with `eslint-config-standard` and prettierx's `spaceBeforeFunctionParen` property :tada:  

It IS standard :tada: but more native

Shinigami92:
How to use this when testing locally:

```shell
nvm use 14 # or 12
rm -Rf node_modules package-lock.json
npm install
npm run build
npm run format
npm run lint
npm run test
```

If you miss anything, please let me know and I will look if I can satisfy your needs

pioardi:
So when I run npm run test, how can I check if you run all the lint and format scripts as we want ? 
I really see a lot of benefits using standard, it is used in so much libraries, including nodejs itself
ts-standard is something that I like less to be honest I do not see it so mature and widely used

Shinigami92:
We could add the lint step to the test script
`standard` is just a preset wrapper around `eslint`

ST-DDT (approved):

.eslintrc.js
```diff
@@ -0,0 +1,15 @@
+module.exports = {
+  env: {
+    es2021: true,
+    node: true,
+    mocha: true
+  },
+  extends: ['standard'],
```
jerome-benoit, line 7: is prettier configuration inside standard?

.eslintrc.js
```diff
@@ -0,0 +1,15 @@
...
+  parser: '@typescript-eslint/parser',
+  parserOptions: {
+    ecmaVersion: 12,
+    sourceType: 'module'
+  },
+  plugins: ['@typescript-eslint'],
```
jerome-benoit, line 13: is prettierx plugin enabled by default?

.eslintrc.js
Shinigami92: I think yes, because we don't use `prettier` here but `prettierx` which is configured to perform like `standard`

I already ran `npm run format` on this PR, so you can see the diff what changes when performing npm run format + npm run lint

.eslintrc.js
Shinigami92: I think so

Shinigami92:
@pioardi @jerome-benoit 

I ran `npx eslint --init` and this is the output:

Screenshot_20210208_185731

So you can see I selected the `standard` style-guide

pioardi:
> We could add the lint step to the test script
> `standard` is just a preset wrapper around `eslint`

I don't know if tun the lint step will fix the problem really, we need to check if during a new implementation anyone used the linter that we provide

jerome-benoit:
> > We could add the lint step to the test script
> > `standard` is just a preset wrapper around `eslint`
> 
> I don't know if tun the lint step will fix the problem really, we need to check if during a new implementation anyone used the linter that we provide

A lint run can done on PR also by using GitHub CI, GitHub highlight lint failures then.

.eslintrc.js
Shinigami92: Can you check the config now

.eslintrc.js
```diff
@@ -0,0 +1,37 @@
+module.exports = {
+  env: {
+    es2021: true,
```
Shinigami92, line 3: @jerome-benoit What do you think about this flag? It was added by the init script :thinking:

.github/workflows/ci.yml
```diff
@@ -29,13 +29,16 @@ jobs:
       - name: Build
         run: npm run build --if-present
 
+      - name: Lint
+        run: npm run lint
```
Shinigami92, line 33: @jerome-benoit @pioardi I have added the lint step here!

.eslintrc.js
jerome-benoit: Hum, is it possible to enable the same es version than the one in tsconfig.json?

.eslintrc.js
```diff
@@ -0,0 +1,37 @@
...
+  parser: '@typescript-eslint/parser',
+  parserOptions: {
+    ecmaVersion: 12,
+    sourceType: 'module'
+  },
+  plugins: ['@typescript-eslint', 'prettierx'],
+  extends: [
+    'standard',
+    'eslint:recommended',
+    'plugin:@typescript-eslint/recommended',
```
jerome-benoit, line 16: I do not remember that one contains `@typescript-eslint/recommended-requiring-type-checking ` ...

jerome-benoit (dismissed):
Look good to me. Refinement can always be done later one anyway.

.eslintrc.js
Shinigami92: `Environment key "es2018" is unknown` ... no

jerome-benoit: Let's suppose the default configured values are sensible ones and keep them ;)

.eslintrc.js
Shinigami92: Seems like setting `plugin:@typescript-eslint/recommended-requiring-type-checking` in the current stage with a `tsconfi.lint.json` produces to much errors right now, because also tests and other js files are affected by this

I would vote for creating another branch for that later

Shinigami92:
In a later followup branch, we can use `plugin:@typescript-eslint/recommended-requiring-type-checking` instead of `plugin:@typescript-eslint/recommended` and use following `tsconfig.lint.json`

```json
{
  "compilerOptions": {
    "target": "ES2018",
    "module": "CommonJS",
    "rootDir": ".",
    "esModuleInterop": true,
    "noEmit": true,
    "strict": true
  },
  "exclude": ["node_modules"]
}
```

Also we need to set `parserOptions.project: ['./tsconfig.lint.json']`

And then we need to add JSDoc to test files or even better: also convert the tests to TS :thinking:  
Sadly we cannot do this in this PR, because it produces to many errors in all the js files

jerome-benoit (dismissed):

.github/workflows/ci.yml
pioardi: This is the reason why I see warnings on this file indicating es-lint issues ? I would love that oh yeah

Shinigami92: Yes! This two lines provide you with warnings and errors in the `files changes` tab

.eslintrc.js
```diff
@@ -0,0 +1,37 @@
...
+    'plugin:prettierx/standardx',
+    'plugin:prettierx/@typescript-eslint'
+  ],
+  rules: {
+    'no-void': 'off',
+
+    // Disabled because we intentionally have any for generics
+    '@typescript-eslint/no-explicit-any': 'off',
```
jerome-benoit, line 24: I really prefer to have that one disabled at a line level is TS files. 
One of the purpose of TS is to have strong typing, so using 'any' shall remain a documented exception in the code.

Shinigami92: Yeah, okay I will bring this one to line level

.eslintrc.js
```diff
@@ -0,0 +1,37 @@
...
+  plugins: ['@typescript-eslint', 'prettierx'],
+  extends: [
+    'standard',
+    'eslint:recommended',
+    'plugin:@typescript-eslint/recommended',
+    'plugin:prettierx/standardx',
+    'plugin:prettierx/@typescript-eslint'
+  ],
+  rules: {
+    'no-void': 'off',
```
jerome-benoit, line 21: I'm still having a debate with myself to see if that's really worth putting that one globally ...

.eslintrc.js
```diff
@@ -0,0 +1,37 @@
...
+
+    // Disabled because we intentionally have any for generics
+    '@typescript-eslint/no-explicit-any': 'off',
+
+    // Disabled because it reports these for js files
+    '@typescript-eslint/no-var-requires': 'off',
```
jerome-benoit, line 27: I think that one could be easily removed by fixing the JS files.

Shinigami92: Yes, but I would prefer (that someone) do this in a separat PR

.github/workflows/ci.yml
pioardi: This definitevely remove any doubt for me about those changes, great !

pioardi: And then do you think that we should fix those warnings in this PR or into another one ?

Shinigami92: This PR is already huge, lets tackle these in another PR
Because they are in `examples` and `benchmarks`, maybe it would be nice if you can tackle them, so we can split the work a bit
But if you don't have time, I also could tackle these later on

pioardi: Sounds good but please do that as soon as we can so that when some PR comes we are able to see if there is any warning due to the changes

pioardi (approved):
Is good to go for me. Nice job we still have a standard code base but we fixed some problems that we had with ts-standard.

@jerome-benoit  let us know if for you is good to go

.eslintrc.js
Shinigami92: It conflicts with [@typescript-eslint/no-floating-promises](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/no-floating-promises.md)

jerome-benoit:
> Is good to go for me. Nice job we still have a standard code base but we fixed some problems that we had with ts-standard.
> 
> @jerome-benoit let us know if for you is good to go

I've already approved that PR ;)

PR #86, opened 2021-02-08
Rename functions and methods to not use prefix underscore

### LINKED ISSUE #75: Rename functions and methods to not use prefix underscore

Shinigami92:
closes #75


pioardi (approved):
is good to go for me

Shinigami92:
@pioardi Should I also update the Changelog in this PR? Or should we do this shortly before packaging up a release?

Also is there anything else that we need to update in this PR?

Do you will note it on a peace of paper on your test for TODO? :D

pioardi:
Oh yeah, a note to the change log would be great.
No paper lol

Shinigami92:
Mh, did users already needed to use `const { FixedThreadPool } = require("poolifier");`

So maybe it's only a breaking change if they abstractly used something like this `const { FixedThreadPool } = require("poolifier/lib/fixed");`
But I thing they should, do that.

I would suggest to write a Changelog when everything is ready for a release
So we can look what really has changed ^^

pioardi:
mmmm would be much more complex and error prone to do that at release time, maybe we should start to add a line into the change log for any PR ?

tests/fixed.test.js
```diff
@@ -17,7 +17,7 @@ describe('Fixed thread pool test suite ', () => {
   it('Choose worker round robin test', async () => {
     const results = new Set()
     for (let i = 0; i < numThreads; i++) {
-      results.add(pool._chooseWorker().threadId)
+      results.add(pool.chooseWorker().threadId)
```
pioardi, line 20: If choose worker is protected how we are accessing it here ? mmmmmm

pioardi:
I vote to add this to our change log :

_chooseWorker, _newWorker, _execute , _chooseWorker, _checkAlive, _run , _runAsync methods are no more accessible

I also put one comment to understand how we using the chooseWorker method from our tests since it is protected

tests/fixed.test.js
Shinigami92: member accessors are just a TS feature that will be removed while transpiling
JS is pure public (not anymore with `#`-private properties, but please don't use them)

pioardi: Ok so in our change log we should not say that are no more accessible but that the _ is removed and we anyway are not intended to be used

CHANGELOG.md
```diff
@@ -5,6 +5,37 @@ All notable changes to this project will be documented in this file.
...
+const { DynamicThreadPool } = require("poolifier/lib/dynamic");
+```
+
+But you should always prefer just using
+
+```js
+const { DynamicThreadPool } = require("poolifier");
+```
+
+#### Internal (protected) methods has renamed
```
pioardi, line 29: I would add "Those methods are not intended to be used from final users"

pioardi (approved):
Good to go

Shinigami92:
@pioardi I intentionally used a squash-merge here

Now you can see how clean it looks in the history

image

pioardi:
I love squash commits, thanks

PR #87, opened 2021-02-08
Sync version in package.json with the one in lock file.

jerome-benoit:


Shinigami92:
Let's merge/update that before we release the v2  
So we don't need to fix that here and there on the road to v2

pioardi:
@jerome-benoit  thank you thank you thank you for all your commitment on this project.
Please let's talk a little on gitter to meet and know you better if you like :) 

Anyway the PR is good to go, just update from the master and we will merge

pioardi (approved):

Issue #66, opened 2020-11-12
Provide a cluster worker pool


Shinigami92:
I have rewritten your `FixedThreadPool` to a `FixedClusterPool` :sparkles:  
When my TypeScript branch (#64) is merged, I will create a new PR to also let poolifier provide a `Process/Cluster Pool` 

If you want, you could give me repo write permissions with the possibility to only create branches but not access or merge into master branch
Then we can work together on this PR together (e.g. write tests)

Shinigami92:
My current code:
(_comments and some variable names need to be changed_)

<details>
<summary>fixed.ts</summary>

```ts
// Christopher will create a PR in poolifier repo
// Therefore this code is temporarily

import { SendHandle } from 'child_process';
import * as cluster from 'cluster';
import { Worker } from 'cluster';

function empty(): void {
    // empty
}

export type Draft<T> = { -readonly [P in keyof T]?: T[P] };

export type WorkerWithMessageChannel = Worker; // & Draft<MessageChannel>;

export interface FixedClusterPoolOptions {
    /**
     * A function that will listen for error event on each worker thread.
     */
    errorHandler?: (this: Worker, e: Error) => void;
    /**
     * A function that will listen for online event on each worker thread.
     */
    onlineHandler?: (this: Worker) => void;
    /**
     * A function that will listen for exit event on each worker thread.
     */
    exitHandler?: (this: Worker, code: number) => void;
    /**
     * This is just to avoid not useful warnings message, is used to set `maxListeners` on event emitters (workers are event emitters).
     *
     * @default 1000
     */
    maxTasks?: number;
}

/**
 * A thread pool with a static number of threads, is possible to execute tasks in sync or async mode as you prefer.
 *
 * This pool will select the worker thread in a round robin fashion.
 *
 * @author [Alessandro Pio Ardizio](https://github.com/pioardi)
 * @since 0.0.1
 */
export class FixedClusterPool<Data = any, Response = any> {
    public readonly workers: WorkerWithMessageChannel[] = [];
    public nextWorker: number = 0;

    // threadId as key and an integer value
    public readonly tasks: Map<WorkerWithMessageChannel, number> = new Map<WorkerWithMessageChannel, number>();

    protected _id: number = 0;

    /**
     * @param numThreads Num of threads for this worker pool.
     * @param filePath A file path with implementation of `ThreadWorker` class, relative path is fine.
[...]

PR #88, opened 2021-02-08
Create an abstract class for pool implementation 

jerome-benoit:
To ease the integration of futur pool implementation.

Future pool implementation will permit to refine the API defined for pool implementation. 


pioardi:
I will try to look into this PR tomorrow, for now I can only say that task on node 14.x version is failing

jerome-benoit:
> I will try to look into this PR tomorrow, for now I can only say that task on node 14.x version is failing

Probably an issue with GitHub CI infrastructure ...

Shinigami92:
@jerome-benoit Would it be possible for you to wait a little bit until I at least start working on #66. Hopefully we can see better what we can extract of logic and types

jerome-benoit:
> @jerome-benoit Would it be possible for you to wait a little bit until I at least start working on #66. Hopefully we can see better what we can extract of logic and types

It can wait for the current PRs to be merged, I will update that PR. 
But it's just the global code structure to follow for pool implementation. So merging it before any work on new pool implementation is a must. The API here is not a step in stone, it will change but the overall structure shall be followed. And the API changes are expected to happen during new pool implementation coding.

Shinigami92:
Yeah, I understand your point, but maybe there are differences between a cluster pool and a worker pool
So let's find these diffs first before defining the abstract pool and then trying to figure out what needs to be moved out of the abstraction

IMO it's better to abstract things afterwards instead of forcing an abstraction
I hope you understand what I mean ^^

jerome-benoit:
> Yeah, I understand your point, but maybe there are differences between a cluster pool and a worker pool
> So let's find these diffs first before defining the abstract pool and then trying to figure out what needs to be moved out of the abstraction
> 
> IMO it's better to abstract things afterwards instead of forcing an abstraction
> I hope you understand what I mean ^^
> 


The other way around also works.
I do not care about the methodology used to reach the same design pattern implementation for pool.

--
Jérôme Benoit aka fraggle
Piment Noir - https://piment-noir.org
OpenPGP Key ID : 27B535D3
Key fingerprint : B799 BBF6 8EC8 911B B8D7 CDBC C3B1 92C6 27B5 35D3

Shinigami92:
> The other way around also works. I do not care about the methodology used to reach the same design pattern implementation for pool.

You should observe my #92 PR, and see the diff between `cluster/fixed` and `thread/fixed`
Also between `cluster-worker` and `thread-worker`
Hope you can then see that we need e.g. different imports from `node` and use different strategies here and there
Yes we can create an interface to force the "API", but we could also try to create an `abstract class` holding the common logic

jerome-benoit:
> > The other way around also works. I do not care about the methodology used to reach the same design pattern implementation for pool.
> 
> You should observe my #92 PR, and see the diff between `cluster/fixed` and `thread/fixed`
> Also between `cluster-worker` and `thread-worker`
> Hope you can then see that we need e.g. different imports from `node` and use different strategies here and there
> Yes we can create an interface to force the "API", but we could also try to create an `abstract class` holding the common logic

The pool abstract class here was meant for that ;)

I close that PR and wait for the dust to settle.

PR #91, opened 2021-02-09
Clean up project

Shinigami92:
This PR is a hodgepodge of some basic changes in the project


README.md
```diff
@@ -1,22 +1,21 @@
...
-Poolifier is used to perform heavy CPU bound tasks on nodejs servers, it implements thread pools ( yes, more thread pool implementations, so you can choose which one fit better for you ) using <a href="https://nodejs.org/api/worker_threads.html#worker_threads_worker_threads">worker-threads </a>.<br>
-With poolifier you can improve your <strong>performance</strong> and resolve problems related to the event loop.<br>
-Moreover you can execute your CPU tasks using an API designed to improve the <strong>developer experience</strong>.
+## Why Poolifier?
 
+Poolifier is used to perform heavy CPU bound tasks on nodejs servers, it implements thread pools (yes, more thread pool implementations, so you can choose which one fit better for you) using [worker-threads](https://nodejs.org/api/worker_threads.html#worker_threads_worker_threads).  
+With poolifier you can improve your **performance** and resolve problems related to the event loop.  
+Moreover you can execute your CPU tasks using an API designed to improve the **developer experience**.
 
+## Contents
```
pioardi, line 17: Compatibility and choose a pool links are no more working now

pioardi (commented):
Some links are not working

README.md
Shinigami92: Links should look like that: https://github.com/pioardi/poolifier/blob/clean-up-project/README.md#contents

Shinigami92: Could you tell me where the links are?

jerome-benoit: > Could you tell me where the links are?

In the table of content: 

+ Choose a pool 
+ Team
+ ... 

I think you can also switch it to md syntax: 

```
##### Table of Contents  
[Headers](#headers)  
[Emphasis](#emphasis)  
...snip...    
<a name="headers"/>
## Headers
```

Shinigami92: Oh... thx, I didn't even noticed that there is a TOC ^^

jerome-benoit (approved):

.eslintrc.js
```diff
@@ -32,14 +32,6 @@ module.exports = {
...
       'error',
       { ignoreProperties: true }
-    ],
-
-    'sort-imports': [
-      'warn',
-      {
-        ignoreMemberSort: true,
-        memberSyntaxSortOrder: ['none', 'all', 'single', 'multiple']
-      }
```
jerome-benoit, line 42: I would like to keep that eslint warning for unsorted imports. The fix is just to run `npm run format`.

Shinigami92: Really? But how do you want to auto-format then natively? What's wrong with the native API sorting?

Shinigami92: `npm run format` doesn't solve this warnings
The configuration doesn't satisfy TS native API ordering

jerome-benoit: The eslint setup for that rule is the same of the sorting ordering expected with the native API. Eslint does not sort any imports with it, it's just warn that it should be done.

Shinigami92: Ok, but then something is configured incorrect

Shinigami92: Oh ... I see you cannot see these warnings in master because they are in #92 



But all these 9 warnings are false positives and conflicts with the native ordering by TS native
I think the problem is, that this rule is only for javascript but not for typescript :confused:  
Maybe we should move it to the js override block?

jerome-benoit: That setup is working:

```
    'sort-imports': [
      'warn',
      {
        ignoreMemberSort: true,
        ignoreDeclarationSort: true
      }
    ]
```

Could you please fix it in that PR?

jerome-benoit (approved):

PR #94, opened 2021-02-09
Do not include src directory in the package.

jerome-benoit:


pioardi (approved):
Approved

PR #92, opened 2021-02-09
Provide a cluster worker pool

Shinigami92:
closes #66

PR #97, opened 2021-02-09
Eslint configuration refinement

jerome-benoit:
+ add import configuration;
+ sync ecma version;
+ add prettier configuration.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -38,8 +38,8 @@ module.exports = {
     {
       files: ['*.js'],
       rules: {
-        '@typescript-eslint/no-var-requires': 'off',
-        '@typescript-eslint/no-unused-vars': 'off'
+        '@typescript-eslint/no-unused-vars': 'off',
+        '@typescript-eslint/no-var-requires': 'off'
       }
     },
     {
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -6,7 +6,7 @@ module.exports = {
   },
   parser: '@typescript-eslint/parser',
   parserOptions: {
-    ecmaVersion: 12,
+    ecmaVersion: 2020,
     sourceType: 'module'
   },
   plugins: ['@typescript-eslint', 'prettierx'],
@@ -14,8 +14,14 @@ module.exports = {
     'standard',
     'eslint:recommended',
     'plugin:@typescript-eslint/recommended',
+    'plugin:import/errors',
+    'plugin:import/warnings',
+    'plugin:import/typescript',
     'plugin:prettierx/standardx',
-    'plugin:prettierx/@typescript-eslint'
+    'plugin:prettierx/@typescript-eslint',
+    'prettier',
+    'prettier/standard',
+    'prettier/@typescript-eslint'
   ],
   rules: {
     'no-void': 'off',
@@ -33,6 +39,12 @@ module.exports = {
       files: ['*.js'],
       rules: {
         '@typescript-eslint/no-var-requires': 'off'
+      }
+    },
+    {
+      files: ['examples/typescript/*.ts'],
+      rules: {
+        'import/no-unresolved': 'off'
       }
     }
   ]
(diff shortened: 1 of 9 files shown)


tsconfig.json
```diff
@@ -1,6 +1,6 @@
 {
   "compilerOptions": {
-    "target": "ES2018",
+    "target": "ES2020",
```
Shinigami92, line 3: This is potentially dangerous :warning:  
We want to support Node 12

.eslintrc.js
```diff
@@ -34,6 +40,12 @@ module.exports = {
       rules: {
         '@typescript-eslint/no-var-requires': 'off'
```
Shinigami92, line 41: Could you also add `'@typescript-eslint/no-unused-vars': 'off',` here as we already did in other PRs

https://github.com/pioardi/poolifier/pull/91/files#diff-e2954b558f2aa82baff0e30964490d12942e0e251c1aa56c3294de6ec67b7cf5R35
https://github.com/pioardi/poolifier/pull/92/files#diff-e2954b558f2aa82baff0e30964490d12942e0e251c1aa56c3294de6ec67b7cf5R35

Or maybe we should merge https://github.com/pioardi/poolifier/pull/91 so we don't need this in every single open PR

Shinigami92 (changes requested):

tsconfig.json
jerome-benoit: That's fine, the latest previous LTS version works flawlessly with that target.

Shinigami92: According to this, we should at least use es2019

https://stackoverflow.com/a/59787575/6897682

Otherwise please share your source

Shinigami92: Also this site https://node.green/#ES2020 says that node 12.10.0 doesn't support most of the es2020 features  
Maybe our transpiled source code isn't affected anyways, but just keep save

.eslintrc.js
```diff
@@ -38,7 +38,8 @@ module.exports = {
     {
       files: ['*.js'],
       rules: {
-        '@typescript-eslint/no-var-requires': 'off'
+        '@typescript-eslint/no-var-requires': 'off',
+        '@typescript-eslint/no-unused-vars': 'off'
```
Shinigami92, line 42: Uh I don't want to get you on your nerves with that, trust me ^^'

But you should swap these two lines, because in the other PRs we also have them this way and `u` is one letter before `v` in alphabetical order
So we wont have merge conflicts for this single line

package.json
```diff
@@ -9,6 +9,7 @@
     "test": "npm run build && nyc mocha --exit --timeout 20000 tests/*test.js ",
     "test:debug": "mocha --inspect-brk --exit tests/*test.js ",
     "coverage": "nyc report --reporter=text-lcov | coveralls",
+    "import-sort": "npx import-sort-cli --write '{src,examples}/**/*.ts{,x}'",
```
Shinigami92, line 12: I already added `prettier-plugin-organize-imports` this should sort the imports by default native typescript API
So we don't need a third party sorting anymore
Also this `import-sort` maybe don't supports TS 3.8 `import type` syntax

Shinigami92 (changes requested):

package.json
jerome-benoit: It does not work with prettierx nor prettier.

Shinigami92: Ah I know what the problem is
We need to write something like `"format": "prettier --write . && prettierx --write ."`

Because prettierx is a little bit behind and doesn't support the installed plugin right now
But when running prettier before we have the correct ordering

Shinigami92 (changes requested):

package.json
Shinigami92: We need `"format": "prettier --loglevel silent --write . && prettierx --write ."`

Shinigami92: I also created an issue https://github.com/brodybits/prettierx/issues/463, so hopefully they support these plugins natively like prettier does

Shinigami92 (approved):

PR #100, opened 2021-02-09
Do not push only share

pioardi:
Just for sharing

### DIFF
--- a/src/worker/cluster-worker.ts
+++ b/src/worker/cluster-worker.ts
@@ -45,8 +45,10 @@ export class ClusterWorker<Data = any, Response = any> extends AsyncResource {
         // here you will receive messages
         // console.log('This is the main thread ' + isMainThread)
         if (this.async) {
+          console.log('async run ' + value.data.f)
           this.runInAsyncScope(this.runAsync.bind(this), this, fn, value)
         } else {
+          console.log('sync run ' + value.data.f)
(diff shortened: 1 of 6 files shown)


src/worker/cluster-worker.ts
```diff
@@ -91,7 +99,9 @@ export class ClusterWorker<Data = any, Response = any> extends AsyncResource {
         this.lastTask = Date.now()
       })
       .catch(e => {
-        cluster.worker.send({ error: e, id: value.id })
+        const err = e instanceof Error ? e.message : e
```
pioardi, line 102: this is the real tip @Shinigami92

src/worker/cluster-worker.ts
```diff
@@ -76,7 +78,13 @@ export class ClusterWorker<Data = any, Response = any> extends AsyncResource {
       cluster.worker.send({ data: res, id: value.id })
       this.lastTask = Date.now()
     } catch (e) {
-      cluster.worker.send({ error: e, id: value.id })
+      console.log('ok we have an error ')
+      const err = e instanceof Error ? e.message : e
```
pioardi, line 82: and also this one

Shinigami92:
You should use draft PRs ^^
image

I will convert it to a draft

But, I think I will look into it tomorrow

PR #101, opened 2021-02-09
Add node TS type definition

jerome-benoit:

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -5,7 +5,8 @@
     "outDir": "lib",
     "esModuleInterop": true,
     "declaration": true,
-    "strict": true
+    "strict": true,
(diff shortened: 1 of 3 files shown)


tsconfig.json
```diff
@@ -5,7 +5,8 @@
     "outDir": "lib",
     "esModuleInterop": true,
     "declaration": true,
-    "strict": true
+    "strict": true,
+    "typeRoots": ["./node_modules/@types"]
```
Shinigami92, line 9: According to this https://www.typescriptlang.org/tsconfig#typeRoots we don't need `typeRoots`, it is already used by default

Shinigami92 (changes requested):

Shinigami92 (approved):

PR #104, opened 2021-02-10
Add benchmark script

jerome-benoit:
+ Fix import;
+ Add workerpool execution time comparaison;
+ Rename worker scripts.

Improve eslint configuration and fix promise handling.

```
xx@xx benchmarks (benchmark|✔) % node myBench.js 
Fixed pool take 943ms to work on 1000 tasks
Dynamic pool take 936ms to work on 1000 tasks
workerpool take 1261ms to work on 1000 tasks
worker threads pool take 10005ms to work on 1000 tasks
```
```
xx@xx poolifier-git (benchmark|✔) % npm run benchmark 

> poolifier@1.2.1 benchmark /Users/I339261/SAPDevelop/poolifier-git
> npm run build && node benchmarks/bench.js


> poolifier@1.2.1 build /Users/I339261/SAPDevelop/poolifier-git
> npm run build:clean && tsc


> poolifier@1.2.1 build:clean /Users/I339261/SAPDevelop/poolifier-git
> rimraf lib

PioardiStaticPool x 4,352 ops/sec ±58.91% (8 runs sampled)
PioardiDynamicPool x 6,048 ops/sec ±38.56% (12 runs sampled)
Fastest is PioardiDynamicPool and PioardiStaticPool
```


benchmarks/bench.js
```diff
@@ -1,18 +1,18 @@
 const Benchmark = require('benchmark')
 const suite = new Benchmark.Suite()
-const FixedThreadPool = require('../lib/fixed')
-const DynamicThreadPool = require('../lib/dynamic')
+const { FixedThreadPool } = require('../lib/index')
+const { DynamicThreadPool } = require('../lib/index')
 const size = 30
 const tasks = 1
```
Shinigami92, line 7:
```suggestion

const LIST_FORMATTER = new Intl.ListFormat('en-US', {
  style: 'long',
  type: 'conjunction'
})
```

benchmarks/bench.js
```diff
@@ -67,6 +77,8 @@ async function test () {
     .on('complete', function () {
       this.filter('fastest').map('name')
       console.log('Fastest is ' + this.filter('fastest').map('name'))
```
Shinigami92, line 79:
```suggestion
      console.log(
        'Fastest is ' +
          LIST_FORMATTER.format(this.filter('fastest').map('name'))
      )
```

benchmarks/bench.js
```diff
@@ -67,6 +77,8 @@ async function test () {
     .on('complete', function () {
       this.filter('fastest').map('name')
```
Shinigami92, line 78:
```suggestion
```

Shinigami92 (changes requested):
Could you add an additional benchmark step to the ci workflow?

Shinigami92:
@jerome-benoit 
Maybe we should move the benchmark ci step behind the test step
does it make sense to benchmark it when the tests are failing?

.github/workflows/ci.yml
```diff
@@ -32,6 +32,9 @@ jobs:
       - name: Lint
         run: npm run lint
 
+      - name: Benchmark
```
pioardi: Do we really want to run benchmark tests at each push ?
This is a thing that I would handle separately with a benchmark triggered manually or only on the master branch push.
Moreover we could need a strategy to push benchmarking results when we npm run lint.

pioardi (commented):
Thanks @jerome-benoit  to submit this, I have only one feedback general shape is good

.github/workflows/ci.yml
jerome-benoit: I do not know if GitHub CI permit to define manually triggered steps. 

I think the benchmark run on PR is valuable, it can be changed to a performance regression test if a sensible threshold is defined that pool implementations should never be up to.

pioardi:
This is what I would like at the end of the game on benchmarking:
https://github.com/fastify/fastify#benchmarks


This is what they do :
https://github.com/fastify/benchmarks/blob/master/.github/workflows/benchmarks.yml


They run benchmarks on a schedule and then publish results into their readme, continuous benchmarking with poolifier that win on everything else 🔥 👊 🏆 

But I mean this PR is just to fix the current benchmark so we can just remove the step on the CI if you want and work on all this stuff later on

jerome-benoit:
I see. 
I can remove the CI step running benchmark and you can create an issue on the target to reach regarding automated benchmarks run ;)

src/workers.ts
```diff
@@ -108,6 +108,7 @@ export class ThreadWorker<Data = any, Response = any> extends AsyncResource {
       .then(res => {
         this.parent?.postMessage({ data: res, id: value.id })
         this.lastTask = Date.now()
+        return res
```
Shinigami92, line 111: Could you explain why and if this return is needed? We don't return the result of the `fn` Promise chain anyways :thinking:

Shinigami92:
It's possible to create manual triggerable jobs in GitHub Actions: https://github.blog/changelog/2020-07-06-github-actions-manual-triggers-with-workflow_dispatch/
But I never used them before and they are relatively new

src/workers.ts
jerome-benoit: It's the eslint plugin promise that enforces the rule promise/always-return, which a more than sensible rule that avoids stalled promise resolution in a promises call chain.
So returning the result is the most sensible way to silence that linter error.

Shinigami92: Ah! this is new, reported by `'promise'` eslint plugin
Okay :shrug:  
@pioardi What do you think about this rule? I will let you decide if we need this or not
IMO it isn't because the `res` will never be used somewhere and is just a pollution to nowhere
I really like technical clean code ^^'

jerome-benoit: The code is correct if you return, it's not correct if not, semantically speaking. 

You can also return an empty function but you have to return or throw to be on the safe side.

pioardi: I have no concern with the current code

Shinigami92: I also don't have any problems with the current (master-branch) code
Also every tests are running successfully on the current code
I would prefer to not add this to the code, because it looks like we have forgotten to use the returned value in any way

Shinigami92: Okay I think at least returning `null` is a compromise

Shinigami92:
@jerome-benoit 
Do you want that I approve this now so you can merge it? Or do you want to try a `workflow_dispatch` in this PR?

jerome-benoit:
> @jerome-benoit
> Do you want that I approve this now so you can merge it? Or do you want to try a `workflow_dispatch` in this PR?

Let me return more null at promise resolution when the result is not used first.

Shinigami92 (dismissed):

pioardi (approved):
Good to go

jerome-benoit:
I let one of you do the merge.

Thanks.

Shinigami92 (approved):

PR #113, opened 2021-02-10
Only package necessary files

Shinigami92:


package.json
```diff
@@ -46,6 +46,9 @@
     "url": "https://github.com/pioardi/poolifier/issues"
   },
   "homepage": "https://github.com/pioardi/poolifier#readme",
+  "files": [
+    "lib"
```
pioardi, line 50: are we sure that this is going to work ? I mean should be lib/* or something like ?

pioardi:
By the way should we add an npm run build to this workflow ? https://github.com/pioardi/poolifier/blob/master/.github/workflows/npmpublish.yml

package.json
Shinigami92: Yes, you can check it via `npm publish --dry-run`

Shinigami92:
> By the way should we add an npm run build to this workflow ? [https://github.com/pioardi/poolifier/blob/master/.github/workflows/npmpublish.yml](https://github.com/pioardi/poolifier/blob/master/.github/workflows/npmpublish.yml?rgh-link-date=2021-02-10T16%3A09%3A51Z)

YES 😬!!!

Shinigami92:
> YES !!!

@pioardi Oh nevermind, I think it's working because the `test` script already has `npm run build` in it

jerome-benoit (approved):

Issue #96, opened 2021-02-09
Improve type definition for Data


Shinigami92:
We could try to use following types to improve the type definition of `Data`

[TS Playground](https://www.typescriptlang.org/play?#code/C4TwDgpgBAUgygeQHIAUBOBLAth4GBu0AvFAHYCuWARhGlAD5RUD2zANhAIakNQDOwTKQDmvCmzYBuAFChIsREgBqnNuWILk6bLgLRG8ZAEE0aTiF6GkCKgCsIAY2Ay50KzftOoJAN5QA2gDWAFz8ghgiALqhVipqEJJQAL4u4G6KJmYWJJnmADyxquoAfDLSDsykAlAAJpzAnACMMYpx6t5QPinlldV1DQBMLchtGgDkwBACYzIVVcC19ZwAzMPKRRoALAOzvQv9nJtrox3iUtLlbJx8fFAAIiCknDgOACoAFmhcNSisbHnSKD3JZQCAAD0mpBqt0K8Q6sPUABpAVAAEpTMC9aDgyHQzTrOEkBEQaTFToooFgchUNgYBygsGOciTAAUB1CdyWAEpQuhmDg+BA8ui+JiqhAyT4KUCZV9gOQ0Dw+QKIAA6L58diEFl+TihRoADkRTH1A2SUGu+JOlpFYsFXJkMuS0iSF1IEAA7vdHs86R8vpwfn8AU7daEKNRaMlkaGLeHKDQ0MaqPHI3RXcUWQ6gA)

So we could e.g. write

```ts
export class DynamicThreadPool<
  Data extends JSONValue = JSONValue,
  Response extends JSONValue = JSONValue
> extends FixedThreadPool<Data, Response>
```

PR #116, opened 2021-02-10
Encapsulate logic of cluster and thread worker/pool

### LINKED ISSUE #111: Encapsulate logic of cluster and thread worker/pool
**Is your feature request related to a problem? Please describe.**
With #66 we provided a new variant of pool.
Now we have very much duplicate code.
We need to find out what could be encapsulated into abstract classes and interfaces.

**Describe the solution you'd like**
- Create a abstract Pool and Worker definition that encapsulates some of the duplicate logic
- Define generic interfaces (e.g. for _InputOptions_) so you can easily exchange a thread pool with a cluster pool and vice versa.

Shinigami92:
build on top of #92  
closes #111


Shinigami92:
@jerome-benoit If you want you can have a first look :smiley:  

I basically extracted just every common logic into `AbstractWorker` and `AbstractPool`

So from there we could thing about to extract methods that needs to be implemented so we can use wrapper code around them

So for example:

- `destroyWorker(worker: Worker): void`
- `isMain(): boolean`
- Move `setupHook` to a protected member method

pioardi:
I Will take a look tomorriw :)

jerome-benoit:
> I Will take a look tomorriw :)

He's too fast ;)

src/pools/abstract-pool.ts
```diff
@@ -0,0 +1,173 @@
...
+   */
+  public execute (data: Data): Promise<Response> {
+    // configure worker to handle message with the specified task
+    const worker = this.chooseWorker()
+    const previousWorkerIndex = this.tasks.get(worker)
+    if (previousWorkerIndex !== undefined) {
+      this.tasks.set(worker, previousWorkerIndex + 1)
+    } else {
+      throw Error('Worker could not be found in tasks map')
+    }
```
jerome-benoit, line 111: You can factor out these 6 lines to a addWorker() method in the abstract class.

pioardi:
> > I Will take a look tomorriw :)
> 
> He's too fast ;)


Haha you too, but he Is super fast😂

src/pools/abstract-pool.ts
```diff
@@ -0,0 +1,173 @@
...
+    port: Worker,
+    listener: (message: MessageValue<Response>) => void
+  ): void
+
+  protected abstract unregisterWorkerMessageListener (
+    port: Worker,
+    listener: (message: MessageValue<Response>) => void
+  ): void
+
+  protected internalExecute (worker: Worker, id: number): Promise<Response> {
```
jerome-benoit, line 128: As a C developer, I'm used to __execute() way of naming for function used internally. 
Is there a similar convention in the TS, JS world?

src/pools/abstract-pool.ts
```diff
@@ -0,0 +1,173 @@
...
+          }
+          if (message.error) reject(message.error)
+          else resolve(message.data as Response)
+        }
+      }
+      this.registerWorkerMessageListener(worker, listener)
+    })
+  }
+
+  protected chooseWorker (): Worker {
```
jerome-benoit, line 147: That one is targeted to become a strategy (a callback with a defined signature) to be choose at context call ;)

Shinigami92:
@pioardi @jerome-benoit Oh sorry, I just was so deep in the working-tunnel, I didn't recognized that you have written some reviews :D  
I will read them now...

This PR is still a Draft :P

src/pools/abstract-pool.ts
Shinigami92: In Vue/Vuetify they use `internal` as prefix
The underscore (`_`) prefix is the more oldish way of writing a private field/method
Nowadays we have `#` as a real private js field/method (it can really not be accessed from outside!)
But in typescript you have the `private` keyword, and so you don't need the `_` prefix

The `__` is one way to solve this, it's interchangeable with `internal`, but reads much more technical :shrug: 
@pioardi what would you prefer? `__` or `internal`?

src/pools/abstract-pool.ts
Shinigami92: Yeah... I already tried to tackle that, BUT in dynamic-*-pool it uses `super.chooseWorker()` 😬

```ts
type IDynamicPool<Worker extends IWorker, Data, Response> = (
  | DynamicClusterPool<Data, Response>
  | DynamicThreadPool<Data, Response>
) &
  AbstractPool<Worker, Data, Response>

export function dynamicWorkerChooser<
  Worker extends IWorker,
  Data,
  Response,
  DynamicPool extends IDynamicPool<Worker, Data, Response>
> (this: DynamicPool, superChooseWorker: () => Worker): Worker {
  // ...
```

And then it could be called e.g. like this:
```ts
return dynamicWorkerChooser.bind(this, () => super.chooseWorker())
```

But when accessing functions like `internalNewWorker` or `sendToWorker` in `this(: DynamicPool)`, it cannot access them, because it tells me that `this` has no method defined like the ones :confused:

Shinigami92:
Okay... a friend wants to play Overwatch :D  
So I'm now finished for today :wave: 

@jerome-benoit Do you want to create a PR on top of this PR to propose a strategy for `chooseWorker` in dynamic pools?  
So I can review this tomorrow.
I currently have not a real idea, just trying.
My idea would work in plain JS, but it's by far not TS safe :confused: 

Good night :wave: :night_with_stars:

src/pools/abstract-pool.ts
```diff
@@ -0,0 +1,172 @@
+import EventEmitter from 'events'
+import type { MessageValue } from '../utility-types'
+
+export type ErrorHandler<Worker> = (this: Worker, e: Error) => void
+export type OnlineHandler<Worker> = (this: Worker) => void
+export type ExitHandler<Worker> = (this: Worker, code: number) => void
+
+export interface PoolOptions<Worker> {
```
pioardi, line 8: We could consider to have more specific options since that worker_thread has some parameters that cluster does not and viceversa.

Is better to keep options specific

src/pools/abstract-pool.ts
```diff
@@ -0,0 +1,172 @@
...
+
+    for (let i = 1; i <= this.numWorkers; i++) {
+      this.internalNewWorker()
+    }
+
+    this.emitter = new PoolEmitter()
+  }
+
+  protected setupHook (): void {
+    // Can be overridden
```
pioardi, line 79: This can be overriden or has to be overidden?
I mean if override is mandatory this should be an abstract method right?

src/pools/abstract-pool.ts
pioardi: I would stay with the internal, we could add the # if it make the methods really private.
But is not mandatory for now in my opinion

src/pools/abstract-pool.ts
```diff
@@ -0,0 +1,172 @@
...
+
+interface IWorker {
+  on(event: 'error', handler: ErrorHandler<this>): void
+  on(event: 'online', handler: OnlineHandler<this>): void
+  on(event: 'exit', handler: ExitHandler<this>): void
+}
+
+class PoolEmitter extends EventEmitter {}
+
+export abstract class AbstractPool<
```
pioardi, line 37: Abstract class is ok to avoid code duplication, but we should add an interface( IPool could be good as name) that only has one method to implement, the execute method.

In this way if we want to implement a pool that is not suitable to extends the Abstract class we can do it and end users still use the same method signature

pioardi:
General shape is good, just few comments

src/pools/abstract-pool.ts
Shinigami92: Yeah it's possible to define types **if needed** later on like 
```ts
export interface FixedThreadPoolOptions extends PoolOptions<ThreadWorkerWithMessageChannel> {
  additionalProperty: boolean;
  // ...
}

File: src/pools/abstract-pool.ts (hunk 5 above)
Comment (Shinigami92) [reply src/pools/abstract-pool.ts:79]: Like the comment say: You **can**  
So the function will called but it's default impl is just empty  
If you need to do something (like `setupMaster`) you can hook it up

File: src/pools/abstract-pool.ts (hunk 6 above)
Comment (Shinigami92) [reply src/pools/abstract-pool.ts:37]: Will add this

File: src/pools/abstract-pool.ts (hunk 4 above)
Comment (pioardi) [reply src/pools/abstract-pool.ts:8]: Sounds good it should work for later, I am not familiar with Typescript but should work fine

Comment (Shinigami92) [reply src/pools/abstract-pool.ts:8]: Example can now be seen via `ClusterPoolOptions` in commit https://github.com/pioardi/poolifier/pull/116/commits/6e19fd2e0a86ec4e3cc78b70cadc76da1e086876

File: src/pools/abstract-pool.ts (hunk 6 above)
Comment (Shinigami92) [reply src/pools/abstract-pool.ts:37]: Exposed IPool interface in [`d00ce3f` (Define IPool interface)](https://github.com/pioardi/poolifier/pull/116/commits/d00ce3f6ae775a86eb3aeb9c645a87c2ce270575)

File: src/pools/pool.ts [hunk 7]
```diff
@@ -0,0 +1,26 @@
...
+  Data = any,
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  Response = any
+> {
+  readonly workers: Worker[]
+  // nextWorker: number // Do we want to define this here?
+  readonly tasks: Map<Worker, number>
+  readonly emitter: PoolEmitter
+  destroy(): Promise<void>
+  execute(data: Data): Promise<Response>
```
Shinigami92, line 25: @pioardi 
I currently exposed **any** public definitions
_(excluding `nextWorker`, because I don't think that we really need to expose this one)_

But should we define `workers`, `tasks` and `emitter`?
Users could also use things like `class MyOwnPool implements AbstractPool<IWorker> {}`

pioardi: We should export only what we want the end user to use.
The AbstractPool is not intended to be exposed for me.

src/pools/pool.ts
```diff
@@ -0,0 +1,26 @@
...
+}
+
+export interface IPool<
+  Worker extends IWorker,
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  Data = any,
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  Response = any
+> {
+  readonly workers: Worker[]
```
pioardi, line 20: this is an internal variable that does not need to be on the interface

src/pools/pool.ts
```diff
@@ -0,0 +1,26 @@
...
+  // nextWorker: number // Do we want to define this here?
+  readonly tasks: Map<Worker, number>
```
pioardi, line 22: this is an internal variable that does not need to be on the interface

src/pools/pool.ts
```diff
@@ -0,0 +1,26 @@
...
+  readonly emitter: PoolEmitter
```
pioardi, line 23: this is an internal variable that does not need to be on the interface

src/pools/pool.ts
Shinigami92: I would suggest to expose it, so users can potentially create their own implementation

But I think I will remove `workers`, `tasks` and `emitter` from `IPool` because the have nothing todo with the usage of the pool, but only its values in `AbstractPool` behinde the scenes  
So if users want to have these they can use `AbstractPool` as interface instead of extending it

Shinigami92:
@pioardi I know we need to extract stuff out of `(Cluster/Thread)Worker constructor` and also `Dynamic(Cluster/Thread)Pool chooseWorker`  
But would it be okay for you that we merge #92 as squash-merge and then this PR as squash-merge now?  
If you want I also could create tickets for the both above things
When we merge it, I could work on #96 now

src/pools/pool.ts
pioardi: So if an end user want to implement a pool would be better that implement it into our project so that is shared with everyone, right?

This is why I prefer to not expose the AbstractPool, do you agree ?

Shinigami92: I personally if we use the friendly way and relay on a positive biased community that are hyped to provide features to our project instead of forcing them and hiding stuff from them just because we can :slightly_smiling_face: 

But you have the last word on that, because you are the main-author of the project -> so yes: expose, or no: not expose

src/pools/abstract-pool.ts
pioardi: @jerome-benoit  do you have any further suggestion on this?

src/pools/pool.ts
pioardi: Oh I would prefer not, but I don't want to have the last word just because I am the author.
So we have 1 vote for yes and 1 vote for no.

@jerome-benoit  your vote will win :)

src/pools/abstract-pool.ts
Shinigami92: @pioardi Let's move on with the discussion in #122 so we don't suffer in chaos :slightly_smiling_face:

src/pools/pool.ts
Shinigami92: @jerome-benoit 



_so I can proceed to work_ :smile:

Shinigami92: @pioardi if @jerome-benoit doesn't see this, should we flip a coin? :laughing:  
Or should we advertise this request-for-vote more to him?
_I want to proceed_ .-.

pioardi: haha @Shinigami92  you are so fast, don't worry. Let's wait 1/2 hours and if he does not see there I will let you know :) 

You are working so much on this, thanks my friend

Shinigami92: I will go play Heroes of the Storm now, will come back after some rounds :)

Shinigami92: Three rounds HotS finished

Shinigami92: @ST-DDT maybe you can help with a vote here :thinking:

pioardi: I don't want to block you, if you can to proceed go with no export for the AbstractPool or wait a bit :D

And yes if @ST-DTT wants to contribute to this project I would be happy to expand the community.
I way that he/she reviewed already some PR

Shinigami92: He is my co-worker personal friend ^^
Okay, I will remove the export as you said and we can discuss this later

Shinigami92: Done https://github.com/pioardi/poolifier/pull/116/commits/de5c6d7d3bb128532ac86b684f6a3eafd56b464b

src/index.ts
```diff
@@ -1,4 +1,4 @@
-export { AbstractPool } from './pools/abstract-pool'
+// export { AbstractPool } from './pools/abstract-pool'
```
pioardi, line 1: Remove also the comment and we are ready :D

Shinigami92: 😏

Shinigami92: Done https://github.com/pioardi/poolifier/pull/116/commits/e82a18ed6f1b871eeb9dc9df254cfad4508b7f2b

Shinigami92:
@pioardi Do you need anything else before approval?

Shinigami92:
Ok, I think it won't go any further today :(
I'll be going to sleep soon to catch up on sleep

src/pools/pool.ts
pioardi: Thanks @Shinigami92 

Your co-worker and friend is welcome on this community as each other :) Collaborator grants will be added only after a while but he can still contribute when he wants :)

pioardi (approved):
Dear @Shinigami92  this PR for me can be merged, we reviewed the changes, all the tests pass, duplicated code is fixed, interface is defined.
I cannot thank you enough for all your work on that.

The PR is approved, based on your feelings I will let you decide if wait for @jerome-benoit  for a while or not :) 

My preference is obviously to wait a while but I understand and appreciate your enthusiasm

src/pools/pool.ts
ST-DDT: He is too fast...^^
I know it's a little late but:

- If it's internal API, then don't expose it.
- If it's a useful base class, then expose it.

In this case, I don't see anything internal in there, so I would "vote" to expose it, unless it is likely to change.
The reason for that is simple, if the base class is not exposed, then other developers won't use it to create their own implementations, which usually results in them not sharing their implementation (because it's not compatible anyway). If you put a rock on the road, then people will take another road instead of climbing over it, unless they have to.

pioardi: Hei :) So yes I understand your position( shini and you). It can make sense to expose it maybe but we will need to document really well all the internal methods and give more guide on how to create a pool then.

I will not oppose further to that, it is ok for me, we can anyway handle this later on

src/pools/abstract-pool.ts
```diff
@@ -0,0 +1,173 @@
...
+  /**
+   * A function that will listen for online event on each worker.
+   */
+  onlineHandler?: OnlineHandler<Worker>
+  /**
+   * A function that will listen for exit event on each worker.
+   */
+  exitHandler?: ExitHandler<Worker>
+  /**
+   * This is just to avoid not useful warnings message, is used to set `maxListeners` on event emitters (workers are event emitters).
```
ST-DDT, line 29:
```sugestion
   * This is just to avoid non-useful warning messages. Will be used to set `maxListeners` on event emitters (workers are event emitters).
```

But IMO this description is lacking some information on which warning messages are considered "non-useful` and what this task do to prevent them/what else happens in that case

src/pools/abstract-pool.ts
```diff
@@ -0,0 +1,173 @@
...
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  Data = any,
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  Response = any
+> implements IPool<Data, Response> {
+  public readonly workers: Worker[] = []
+  public nextWorker: number = 0
+
+  /**
+   * `workerId` as key and an integer value
```
ST-DDT, line 49: `an integer value` doesn't help with understanding what this is used for

src/pools/cluster/dynamic.ts
```diff
@@ -58,11 +51,11 @@ export class DynamicClusterPool<
...
+      const worker = this.internalNewWorker()
+      worker.on('message', (message: MessageValue<Data>) => {
         if (message.kill) {
-          worker.send({ kill: 1 })
-          worker.kill()
+          this.sendToWorker(worker, { kill: 1 })
+          void this.destroyWorker(worker)
           // clean workers from data structures
           const workerIndex = this.workers.indexOf(worker)
           this.workers.splice(workerIndex, 1)
```
ST-DDT, line 61: IMO this cleanup should be part of the destroy code (similar to the `internalNewWorker`), because it operates on the fields of the superclass.

src/pools/abstract-pool.ts
pioardi: So you can attach only a max number of listeners that by default is 10.
https://nodejs.org/api/events.html#events_emitter_setmaxlisteners_n

We can maybe add this documentation to better explain that and also set it to Infinity why not?

Good point

src/pools/abstract-pool.ts
pioardi: Good catch, this was already present it is my bad.
The key is the worker and the value is the number of tasks that has been assigned to that worker since it started.

This is really used only to set to be used on the setMaxEventListeners, I don't remember if is used for other scopes....maybe not so we could potentially removed.

I guess when I added this map I was preparing something for the future like max number of active tasks into a thread( We need to think if we want to add this later on or not).
If you can please create an issue with those points would be great

src/pools/pool.ts
jerome-benoit: > @jerome-benoit
> 
> 
> 
> _so I can proceed to work_ 😄

I'm a bit busy with my paid work currently and I must have a look to the code in that PR to make up a sensible position. 

All I can say right now is that exposing to the outside world abstract class in a thing that I usually not do. I only expose the interface meant to be consumed by others implemented in the class inherited from the abstract class.

jerome-benoit (commented):
The pool abstraction is not properly shared: 
+ each pool type must inherited from the same abstract class;
+ the abstract class implements the base pool type: fixed;
+ fixed pool is then a stub class.

That will permit to more easily add more pool type.
That will permit to implement the strategy design pattern.
That will avoid some empty methods in the abstract class and other 2 level inheritance issues.

src/pools/pool.ts
Shinigami92: Thx for all your points

Lets bring the discussion to #123 (you should copy your texts to there so I don't need to copy your related work in my name)

Shinigami92:
I will merge this now, so we can start working on branching from master, instead of stacking PRs over PRs  
I will also work later on changes that were found by @ST-DDT :+1:

src/pools/cluster/dynamic.ts
Shinigami92: @jerome-benoit Seems to already work on that => https://github.com/pioardi/poolifier/pull/124

Shinigami92:
> The pool abstraction is not properly shared: abstract -> fixed -> dynamic
> 
> * each pool type must inherited from the same abstract class;
> * the abstract class implements the base pool type: fixed;
> * fixed pool is then a stub class.
> 
> That will permit to more easily add more pool type.
> That will permit to implement the strategy design pattern.
> That will avoid some empty methods in the abstract class and other 2 level inheritance issues.

Yes we could/should do this and I (or maybe you?) will create a PR for that and move on  

I will go eat now, make me fresh and have some morning meetings from work  
So I think I will go back on speed-mode in around 2-3 hours (prepare yourself :laughing:)

_nah... I got it, maybe I'm really a bit too fast :sweat_smile:_

PR #120, opened 2021-02-11
Integrate rollup bundler

jerome-benoit:
Add development build with source map for debugging purpose.


rollup.config.mjs
```diff
@@ -0,0 +1,26 @@
...
+export default {
+  input: 'src/index.ts',
+  output: {
+    dir: 'lib',
+    format: 'cjs',
+    sourcemap: !!isDevelopmentBuild,
+    preserveModules: true,
+    preserveModulesRoot: 'src'
+  },
+  external: ['async_hooks', 'events', 'worker_threads'],
```
Shinigami92, line 15: Could you explain why this is needed?  
Do we always need to remember what is used by our tool?  

---

Seems it prints a warning into the console when you remove some externals :thinking:  
Should we try something like this: https://github.com/stevenbenisek/rollup-plugin-auto-external ?

Alternatively IMO it should fail the build if something missing was detected, so we don't build something that is not usable

_sry if I misunderstood this part of rollup, but I personally never used it before. But I heard that it is currently the best lib-bundler :+1: (if you are not relay on `tsc` directly :wink:)_

package.json
```diff
@@ -68,7 +71,8 @@
     "prettier": "^2.2.1",
     "prettier-plugin-organize-imports": "^1.1.1",
     "prettierx": "^0.17.0",
-    "rimraf": "^3.0.2",
+    "rollup": "^2.38.5",
+    "rollup-plugin-delete": "^2.0.0",
     "typescript": "^4.1.4",
```
Shinigami92, line 76: As you already have modified the package(-lock).json in this PR
Could you update `eslint-plugin-promise` and `typescript` and generate updated package-lock?

rollup.config.mjs
jerome-benoit: The build remain usable, missing external modules is just a warning. 
The poolifier code is very lean, it's not necessary to add another rollup plugin to build that external list module, it can maintained manually.

Shinigami92:
Seems you are working on this and it's not finished yet :slightly_smiling_face:  
Because we are working in a team it increase the performance of us all if we use `draft` PRs
So I will convert this to a `draft` PR now

Shinigami92 (approved):

PR #128, opened 2021-02-12
Only allow primitive JSON for transfer between worker and main worker

### LINKED ISSUE #96: Improve type definition for Data

Shinigami92:
closes #96

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,10 @@ But you should always prefer just using
 const { DynamicThreadPool } = require('poolifier')
 ```
 
+#### New type definitions for input data and response
+
+For cluster and thread pools, you can now only send and receive serializable `JSON` data.
+
 #### Internal (protected) methods renaming
 
 Those methods are not intended to be used from final users
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,7 +28,8 @@ const { DynamicThreadPool } = require('poolifier')
 
 #### New type definitions for input data and response
 
-For cluster and thread pools, you can now only send and receive serializable `JSON` data.
+For cluster and thread pools, you can now only send and receive serializable `JSON` data.  
+_This is not a limitation by poolifier but NodeJS._
 
 #### Internal (protected) methods renaming
 
--- a/src/utility-types.ts
+++ b/src/utility-types.ts
@@ -1,6 +1,12 @@
 export type Draft<T> = { -readonly [P in keyof T]?: T[P] }
 
-export interface MessageValue<Data> {
+export type JSONPrimitive = number | boolean | string | null
+// eslint-disable-next-line no-use-before-define
+export type JSONValue = JSONPrimitive | JSONArray | JSONObject
+export type JSONObject = { [k: string]: JSONValue }
+export type JSONArray = Array<JSONValue>
+
+export interface MessageValue<Data extends JSONValue> {
   readonly data?: Data
   readonly id?: number
   readonly kill?: number
--- a/src/utility-types.ts
+++ b/src/utility-types.ts
@@ -6,7 +6,7 @@ export type JSONValue = JSONPrimitive | JSONArray | JSONObject
 export type JSONObject = { [k: string]: JSONValue }
 export type JSONArray = Array<JSONValue>
 
-export interface MessageValue<Data extends JSONValue> {
+export interface MessageValue<Data = unknown> {
   readonly data?: Data
   readonly id?: number
   readonly kill?: number
--- a/package-lock.json
+++ b/package-lock.json
@@ -26,16 +26,16 @@
       "dev": true
     },
     "@babel/core": {
-      "version": "7.12.13",
-      "resolved": "https://registry.npmjs.org/@babel/core/-/core-7.12.13.tgz",
-      "integrity": "sha512-BQKE9kXkPlXHPeqissfxo0lySWJcYdEP0hdtJOH/iJfDdhOCcgtNCjftCJg3qqauB4h+lz2N6ixM++b9DN1Tcw==",
+      "version": "7.12.16",
+      "resolved": "https://registry.npmjs.org/@babel/core/-/core-7.12.16.tgz",
+      "integrity": "sha512-t/hHIB504wWceOeaOoONOhu+gX+hpjfeN6YRBT209X/4sibZQfSF1I0HFRRlBe97UZZosGx5XwUg1ZgNbelmNw==",
       "dev": true,
       "requires": {
         "@babel/code-frame": "^7.12.13",
-        "@babel/generator": "^7.12.13",
+        "@babel/generator": "^7.12.15",
         "@babel/helper-module-transforms": "^7.12.13",
         "@babel/helpers": "^7.12.13",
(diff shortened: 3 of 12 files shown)


Shinigami92:
@pioardi 
I think this theoretically could be a _breaking_ change for users that already use typescript :thinking:  
So I should note add a note to the changelog

I will make a test locally a bit later and report how it feels and how much this improve the usability  

You can also try it by using `yarn link` (or its npm equivalent) and create a project and the use `yarn link "poolifier"` to override your dependency to use your local copy/repo of poolifier

pioardi:
mmmmm so I need to think about those changes( regardless the breaking change).
We really want this restriction I mean we could just advice people that worker_thread and/or cluster processes does not support some types of input mmmmmmmm.....need to think about that

Shinigami92:
> mmmmm so I need to think about those changes( regardless the breaking change).
> We really want this restriction I mean we could just advice people that worker_thread and/or cluster processes does not support some types of input mmmmmmmm.....need to think about that

When I first used poolifier with a colleague at adSoul, we tried to move a non-serializable object between worker and main worker.  
Of course it didn't work, but by then it was already "too late".  
We just thought wrong and didn't look in the node documentation. #RTFM
If we had direct typing, it would have been noticed immediately.

That's the reason why I want to type it :slightly_smiling_face:

src/utility-types.ts
```diff
@@ -1,6 +1,12 @@
 export type Draft<T> = { -readonly [P in keyof T]?: T[P] }
 
-export interface MessageValue<Data> {
+export type JSONPrimitive = number | boolean | string | null
```
pioardi, line 3: we are sure that there are no other JSON primitives that we want to list here?

pioardi:
> > mmmmm so I need to think about those changes( regardless the breaking change).
> > We really want this restriction I mean we could just advice people that worker_thread and/or cluster processes does not support some types of input mmmmmmmm.....need to think about that
> 
> When I first used poolifier with a colleague at adSoul, we tried to move a non-serializable object between worker and main worker.
> Of course it didn't work, but by then it was already "too late".
> We just thought wrong and didn't look in the node documentation. #RTFM
> If we had direct typing, it would have been noticed immediately.
> 
> That's the reason why I want to type it 🙂

Make sense make sense, my concern is more to avoid any too much restrictive type( I also added a comment).

src/utility-types.ts
Shinigami92: Which are you thinking of?

pioardi: Nothing specific but you know types can be more and more and if we miss any here we will cause bugs.
This is my concern

Shinigami92: There are only a limited number of primitive types
The latest (`bigint`) is so new that few scientists use it for big calculations

https://developer.mozilla.org/en-US/docs/Glossary/Primitive

> There are 7 primitive data types: string, number, bigint, boolean, undefined, symbol, and null.

- `string`, `number`, `boolean` and `null` are used
- `undefined` is covered by the indexed object `JSONObject`
- `symbol` cannot be serialized as JSON values
- `bigint` could be added but also this is _currently_ [not supported](https://golb.hplar.ch/2019/01/js-bigint-json.html) by JSON :warning:  

... so indeed we have every JSON primitiv covered

pioardi: mmmmmmmmmmmmmmm @jerome-benoit  what do you think about this?
@Shinigami92 I think is really useful to define this and document it in JSDoc, but we should not apply that to our interface but on the concrete implementations, this is because we don't know if in future some other pool will support any type

Shinigami92: That indeed makes sense! So I will remove it from the abstract Pool and Worker.  
I will try if `unknown` can be applied for the abstraction, so we don't use the evil `any` type

pioardi: Great, I guess we can proceed I think this is beneficial for users

pioardi (dismissed):
Thanks, this help end users to avoid using inputs that are not supported by node js engine

Shinigami92:
> Thanks, this help end users to avoid using inputs that are not supported by node js engine

@pioardi Just a thing for approvals: you should approve a PR when it's out of draft && the author requested your review
Otherwise when pushing additional commits, the approval gets stale and I need to re-request your approval :disappointed:  

---

Now you can approve so I can merge :smiley:

CHANGELOG.md
```diff
@@ -26,6 +26,10 @@ But you should always prefer just using
 const { DynamicThreadPool } = require('poolifier')
 ```
 
+#### New type definitions for input data and response
+
+For cluster and thread pools, you can now only send and receive serializable `JSON` data.
```
pioardi, line 31: I would add that this is due to a nodejs limitation and does not depends by this library.

pioardi (approved):
Good ti go

src/utility-types.ts
jerome-benoit: > That indeed makes sense! So I will remove it from the abstract Pool and Worker.
> I will try if `unknown` can be applied for the abstraction, so we don't use the evil `any` type
> 
> Edit: done

You can also use Record<string, unknown> to ensure a key/value structure is required.

src/utility-types.ts
```diff
@@ -1,6 +1,12 @@
...
+// eslint-disable-next-line no-use-before-define
+export type JSONValue = JSONPrimitive | JSONArray | JSONObject
+export type JSONObject = { [k: string]: JSONValue }
```
jerome-benoit, line 6: The type shortcut is better: Record<string, JSONValue>

Shinigami92: This is NOT possible, because with that you get a reference loop

PR #139, opened 2021-02-13
Add eslint-plugin-spellcheck

Shinigami92:
## PR Checklist

- [ ] Yes
- [x] No

---

I added `eslint-plugin-spellcheck` to check correct wording


Shinigami92:
As you can see here: https://github.com/pioardi/poolifier/runs/1893209690#step:6:10

I let pass a run on this first commit to show that there are indeed some warnings
These will also show up in a review in the future, if occure

src/pools/abstract-pool.ts
```diff
@@ -103,12 +103,12 @@ export abstract class AbstractPool<
...
    * Constructs a new poolifier pool.
    *
-   * @param numWorkers Number of workers that this pool should manage.
+   * @param numberOfWorkers Number of workers that this pool should manage.
    * @param filePath Path to the worker-file.
    * @param opts Options for the pool. Default: `{ maxTasks: 1000 }`
    */
   public constructor (
-    public readonly numWorkers: number,
+    public readonly numberOfWorkers: number,
```
pioardi, line 111: This is not backward compatible, right ?

src/pools/cluster/fixed.ts
```diff
@@ -36,16 +36,16 @@ export class FixedClusterPool<
...
    * Constructs a new poolifier fixed cluster pool.
    *
-   * @param numWorkers Number of workers for this pool.
+   * @param numberOfWorkers Number of workers for this pool.
    * @param filePath Path to an implementation of a `ClusterWorker` file, which can be relative or absolute.
    * @param opts Options for this fixed cluster pool. Default: `{ maxTasks: 1000 }`
    */
   public constructor (
-    numWorkers: number,
+    numberOfWorkers: number,
```
pioardi, line 44: same question as above

src/pools/abstract-pool.ts
Shinigami92: Yes, I just thought I were safe because all tests passes :thinking:  
So I will add a note for that in the changelog

src/pools/cluster/fixed.ts
pioardi: Spell check is a good thing but we should keep backward compatibility as best as we can and I would prefer to not introducing anything that is not reaaaally beneficial if it breaks compatibility

Shinigami92: This is not a breaking change, because
1. its the v2 Cluster Pool
2. its a constructor parameter name

src/pools/abstract-pool.ts
Shinigami92: Oh... I even didn't noticed that VSCode refactoring also changed it in the tests :sweat_smile:

Shinigami92: @pioardi I added backward compatibility in https://github.com/pioardi/poolifier/pull/139/commits/d66606eb9bdf359853779b181598cd7bddb0664a

Shinigami92:
@pioardi Ready for a new review :slightly_smiling_face:

.vscode/settings.json
```diff
@@ -0,0 +1,21 @@
...
+    "lcov",
+    "loglevel",
+    "markdownlint",
+    "piment",
+    "pioardi",
+    "poolifier",
+    "prettierx",
+    "serializable",
+    "standardx",
+    "wessberg",
```
jerome-benoit, line 18: I think you do not need to one anymore.

src/pools/abstract-pool.ts
```diff
@@ -122,15 +122,26 @@ export abstract class AbstractPool<
...
 
     this.emitter = new PoolEmitter()
   }
 
   /**
-   * Setup hook that can be overridden by a Poolifer pool implementation
+   * Number of workers that this pool should manage.
+   *
+   * @returns Number of workers that this pool manages.
+   * @deprecated Please use `numberOfWorkers`. Will be removed in a later version. Only here for reasons of backward compatibility.
```
jerome-benoit, line 136: 'Only here for backward compatibility' should be enough.

jerome-benoit (approved):
Looks good to me.

CHANGELOG.md
```diff
@@ -31,6 +31,10 @@ const { DynamicThreadPool } = require('poolifier')
 For cluster and thread pools, you can now only send and receive serializable `JSON` data.  
 _This is not a limitation by poolifier but NodeJS._
 
+#### Public properties renaming
+
+- Thread Pool's `numWorkers` is now `numberOfWorkers`
```
pioardi, line 36: Please we should indicate how old code looks like from an end user perspective and how  the new should look.
Please indicate that this change is into the constructor

Shinigami92: Seems you misunderstood one thing.
The one part was just a constructor parameter, its not a change for the user at all!
The other one is the Property that you can access like `myPool.numWorkers` and it's now `myPool.numberOfWorkers`
The user will find this directly without any problems

Also there were not any examples in the example folder that provides an example for exact using this property.
So it's just for advances users and they know what to do

PR #161, opened 2021-02-15
Better handling for maxInactiveTime on dynamic pools

pioardi:
## PR Checklist

- [x] Please add a description of your changes.
- [x] We need your changes to come with unit tests in order to keep this project in quality and easy to maintain.
- [x] Please add a link to the open issue or task that this pull request will resolve.
- [x] Add a note to the changelog to indicate the change.

## Does this PR introduce a breaking change?

- [x] Yes
- [ ] No

If yes please indicate it and migration info into the CHANGELOG.md file.

---

Fixes #70 , now a thread/process is not delete when the task submitted take more time than maxInactiveTime configured.
I tested those changes with TDD writing a test that was failing previously and after add the code is not failing so this should cover well this scenario.
After this fix we really really need to work on emit of worker ready and worker exit to built more resilient tests :) 

I will take a look at my own changes better tomorrow, but seems to be already working :)

PR #169, opened 2021-02-16
Improvements for #161 

Shinigami92:
Improvements for #161


tests/worker/cluster/errorWorker.js
```diff
@@ -8,5 +8,5 @@ function error (data) {
 module.exports = new ClusterWorker(error, {
   maxInactiveTime: 500,
   async: false,
-  killBehavior: killBehaviorTypes
```
Shinigami92, line 11: This was buggy before I started working on the improvements

tests/worker/cluster/asyncErrorWorker.js
```diff
@@ -13,5 +13,5 @@ async function error (data) {
 module.exports = new ClusterWorker(error, {
   maxInactiveTime: 500,
   async: true,
-  killBehavior: killBehaviorTypes
```
Shinigami92, line 16: This was buggy before I started working on the improvements

src/worker/worker-options.ts
```diff
@@ -1,24 +1,44 @@
-import type { KillBehavior } from '../utility-types'
-
 /**
- * Kill behavior enumeration
+ * Enumeration of kill behaviors.
  */
-export const killBehaviorTypes = Object.freeze({
+export const KillBehaviors = Object.freeze({
   SOFT: 'SOFT',
```
Shinigami92, line 5: We could also write JSDoc to these properties :thinking:

src/utility-types.ts
```diff
@@ -43,7 +43,7 @@ export interface MessageValue<
   /**
    * Kill code.
    */
-  readonly kill?: KillBehavior | number
+  readonly kill?: KillBehavior | 1
```
Shinigami92, line 46: I think I can remove the `1` because
1. we already changed the default
2. it doesn't survive the check as before :warning: 

> const isKillBehaviorOptionHard = message.kill === killBehaviorTypes.HARD
> `isKillBehaviorOptionHard` was `false` if `message.kill` contains `1` :warning:

Shinigami92:
@pioardi We have a check in abstract worker

```ts
// ...
else if (value.kill) { // This would survive 1, 'HARD' and 'SOFT'
  // Here is time to kill this worker, just clearing the interval
  if (this.interval) clearInterval(this.interval)
  this.emitDestroy()
}
```

What should we do here?

pioardi:
> @pioardi We have a check in abstract worker
> 
> ```ts
> // ...
> else if (value.kill) { // This would survive 1, 'HARD' and 'SOFT'
>   // Here is time to kill this worker, just clearing the interval
>   if (this.interval) clearInterval(this.interval)
>   this.emitDestroy()
> }
> ```
> 
> What should we do here?

Yes it should is the command that is send from the main thread/process to the worker to clear the interval and destroy the async context

src/utility-types.ts
pioardi: 1 is still needed since that we use it to send a message back to the worker to clearInterval and destroy the async context.
This can be changed later we don't need to do everything now :)

Shinigami92:
> > else if (value.kill) { // This would survive 1, 'HARD' and 'SOFT'
> 
> Yes it should is the command that is sent from the main thread/process to the worker to clear the interval and destroy the async context

A okay, so nevertheless what was in the `kill` -> perform the kill-thing?

src/worker/worker-options.ts
pioardi: mmm I don't know, I document the behavior for both on the worker options

pioardi:
> > > else if (value.kill) { // This would survive 1, 'HARD' and 'SOFT'
> > 
> > 
> > Yes it should is the command that is sent from the main thread/process to the worker to clear the interval and destroy the async context
> 
> A okay, so nevertheless what was in the `kill` -> perform the kill-thing?

Yes was just a "signal" to take kill actions, we can change it later using another property that is not kill

src/worker/worker-options.ts
Shinigami92: I want to JSDoc everything that can be a tooltip in an IDE!

image

PR #179, opened 2021-02-17
Example tips

pioardi:
This is the trick that I am suggesting.
I am also thinking anyway that we could also move everything to the abstract pool and have a private property ( isDynamic ) and based on that we can choose workers statically or dynamically ( this is a different solution from the actual one but I wanted to point that out now so we can reflect and decide together)

@jerome-benoit @Shinigami92


Shinigami92:
Wait... :eyes: why does TS build this :eyes: 
This may be an improvement on top of my suggestion
I will try to apply it, give me a second

Shinigami92:
I applied it and it works :eyes: 
Wow thanks, I learned a new thing about TypeScript via this PR :tada: 
Thank you for showing this
I didn't expect that this is possible

pioardi:
It is more JavaScript than TypeScript :)

PR #181, opened 2021-02-17
Issue 160 ci unit test

pioardi:
## PR Checklist

- [x] We need your changes to come with unit tests in order to keep this project in quality and easy to maintain.
- [x] Please add a link to the open issue or task that this pull request will resolve.

- [ ] Yes
- [x] No

---

I fixed the issue number 160, I also added the parallel option to our tests so now each test file will run in parallel from others

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
     "benchmark": "npm run build && node -r source-map-support/register benchmarks/bench.js",
     "benchmark:debug": "npm run build && node -r source-map-support/register --inspect benchmarks/bench.js",
     "benchmark:prod": "npm run build:prod && node -r source-map-support/register benchmarks/bench.js",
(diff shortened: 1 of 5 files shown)


pioardi:
I see now another error on the tests when running the CI, it does not happen locally

pioardi:
The new error is EPIPE, it is happening now rarely compared to the previous but it is more difficult to fix since that we do not have a way to know if the clear interval already happened, we can try to use process.nextTick to fix also the new error

jerome-benoit (approved):
Bug definitely fixed or not, the code consolidation make sense.

pioardi:
> Bug definitely fixed or not, the code consolidation make sense.

Yes also now we send the kill to the worker also for fixed threads that is a good thing.
I will merge this for now and see later how to fix the EPIPE ( is more difficult )

PR #191, opened 2021-02-18
Dynamic change for worker choice algorithms

jerome-benoit:
## PR Checklist

- [ ] Yes
- [x] No

---

+ Fix the pool interface definition hierarchy: use a public interface and a private interface for pool. 
+ Encapsulate worker choice algos inside strategy classes and implement the strategy design pattern for worker choice algos. 


pioardi:
Looks good to me

src/pools/abstract-pool.ts
```diff
@@ -100,9 +100,9 @@ class PoolEmitter extends EventEmitter {}
  */
 export abstract class AbstractPool<
   Worker extends IWorker,
-  Data = unknown,
-  Response = unknown
-> implements IPool<Data, Response> {
+  Data extends JSONValue = JSONValue,
```
Shinigami92, line 103: @pioardi told me that we specially don't want that on `AbstractPool`

src/pools/abstract-pool.ts
```diff
@@ -135,6 +135,11 @@ export abstract class AbstractPool<
    */
   protected nextMessageId: number = 0
 
+  protected workerChoiceCallback: (
+    poolReference: IPool<Worker, Data, Response>,
+    ...args: unknown[]
```
Shinigami92, line 140: We could try to type this with `...T`
its new in TS since 4.0: https://devblogs.microsoft.com/typescript/announcing-typescript-4-0/#variadic-tuple-types

src/pools/cluster/fixed.ts
```diff
@@ -2,6 +2,7 @@ import { fork, isMaster, setupMaster, Worker } from 'cluster'
 import type { JSONValue, MessageValue } from '../../utility-types'
 import type { PoolOptions } from '../abstract-pool'
 import { AbstractPool } from '../abstract-pool'
+import { IPool } from '../pool'
```
Shinigami92, line 5: We should to add `type` here
```suggestion
import type { IPool } from '../pool'
```

src/pools/dynamic-pool.ts
```diff
@@ -1,5 +1,6 @@
 import type { JSONValue } from '../utility-types'
-import type { AbstractPool, IWorker } from './abstract-pool'
+import type { IWorker } from './abstract-pool'
+import { IPool } from './pool'
```
Shinigami92, line 3:
```suggestion
import type { IPool } from './pool'
```

src/pools/pool.ts
```diff
@@ -1,10 +1,17 @@
+import { JSONValue } from '../utility-types'
```
Shinigami92, line 1:
```suggestion
import type { JSONValue } from '../utility-types'
```

src/pools/pool.ts
```diff
@@ -1,10 +1,17 @@
...
+import { AbstractPool, IWorker } from './abstract-pool'
```
Shinigami92, line 2: import type { IWorker } from './abstract-pool'
import { AbstractPool } from './abstract-pool'

src/pools/selection-strategies.ts
```diff
@@ -2,34 +2,25 @@ import type { JSONValue } from '../utility-types'
 import { isKillBehavior, KillBehaviors } from '../worker/worker-options'
 import type { IWorker } from './abstract-pool'
 import type { IDynamicPool } from './dynamic-pool'
+import { IPool } from './pool'
```
Shinigami92, line 5:
```suggestion
import type { IPool } from './pool'
```

src/pools/selection-strategies.ts
```diff
@@ -61,7 +52,7 @@ export function findFreeWorkerBasedOnTasks<Worker> (
  * @param createAndSetupWorker `createAndSetupWorker` bounded function.
  * @param registerWorkerMessageListener `registerWorkerMessageListener` bounded function.
  * @param destroyWorker `destroyWorker` bounded function.
- * @returns The chosen one.
+ * @returns The chosen worker.
```
Shinigami92, line 55: :cry: I hoped I could provide an easter egg :laughing:

src/pools/thread/fixed.ts
```diff
@@ -2,6 +2,7 @@ import { isMainThread, MessageChannel, SHARE_ENV, Worker } from 'worker_threads'
 import type { Draft, JSONValue, MessageValue } from '../../utility-types'
 import type { PoolOptions } from '../abstract-pool'
 import { AbstractPool } from '../abstract-pool'
+import { IPool } from '../pool'
```
Shinigami92, line 5:
```suggestion
import type { IPool } from '../pool'
```

Shinigami92 (changes requested):

src/pools/abstract-pool.ts
pioardi: Yes I would prefer to have it much more open

src/pools/selection-strategies.ts
pioardi: lol

jerome-benoit: I can keep it ;)

src/pools/cluster/fixed.ts
```diff
@@ -29,10 +30,9 @@ export interface ClusterPoolOptions extends PoolOptions<Worker> {
  * @author [Christopher Quadflieg](https://github.com/Shinigami92)
  * @since 2.0.0
  */
-export class FixedClusterPool<
-  Data = unknown,
-  Response = unknown
-> extends AbstractPool<Worker, Data, Response> {
+export class FixedClusterPool<Data = unknown, Response = unknown>
+  extends AbstractPool<Worker, Data, Response>
+  implements IPool<Worker, Data, Response> {
```
Shinigami92, line 35: We don't need the `implements IPool here` because it's already in `AbstractPool`

src/pools/pool.ts
```diff
@@ -1,10 +1,17 @@
...
+
 /**
  * Contract definition for a poolifier pool.
  *
  * @template Data Type of data sent to the worker. This can only be serializable data.
  * @template Response Type of response of execution. This can only be serializable data.
  */
-export interface IPool<Data = unknown, Response = unknown> {
+export interface IPool<
+  Worker extends IWorker,
```
Shinigami92, line 11: `IPool` doesn't make use of `Worker` so we don't need it here

src/pools/pool.ts
```diff
@@ -1,10 +1,17 @@
...
+  Data = unknown,
+  Response = unknown
+> extends AbstractPool<Worker, Data, Response> {
```
Shinigami92, line 14: Wait what?! Why do you want to let an interface extend from an abstract class?
The abstract class is the implementation of an interface that provides default / basic stuff

src/pools/selection-strategies.ts
Shinigami92: I would like to keep it :smiley:

Shinigami92 (changes requested):
You are currently changing really much in this PR that should just add / redesign the worker choice via a callback :thinking:  
Could you please revert most of your unrelated changes?
In the current version I cannot agree to merge it

src/pools/cluster/fixed.ts
jerome-benoit: A pool implements the interface pool interface, whatever it extends.

src/pools/pool.ts
jerome-benoit: It's needed to allow to properly reuse the same typed implementation for all callbacks on worker-threads or cluster pool, dynamic of fixed.

src/pools/pool.ts
jerome-benoit: It's exactly the same trick you've used with the IDynamicPool interface definition you've made to make the poolReference use inside the dynamic choice functions works. I do not really like but to code and test the initial callbacks registration, that's enough.  

Since the goal here is to make an API for callbacks that can be reused for all pool types, all worker type, a proper internal interface must be done to allow a transparent pool reuse inside the callbacks.

Shinigami92:
<details>
  <summary>Jerome's Class Diagram</summary>

  ```plantuml
  @startuml
    title Jerome's Class Diagram
  
    interface IPoolInternal<Worker, Data, Response> implements AbstractPool
    interface IPool<Data, Response>
  
    abstract class AbstractPool<Worker, Data, Response> implements IPool, IPoolInternal
  
    class FixedThreadPool<Data, Response> extends AbstractPool implements IPoolInternal
    class DynamicThreadPool<Data, Response> extends FixedThreadPool implements IPoolInternal
  
    class FixedClusterPool<Data, Response> extends AbstractPool implements IPoolInternal
    class DynamicClusterPool<Data, Response> extends FixedClusterPool implements IPoolInternal
  @enduml
  ```
</details>

<details>
  <summary>Shinigami's Class Diagram</summary>

  ```plantuml
  @startuml
    title Shinigami's Class Diagram

    interface IPool<Data, Response>

    abstract class AbstractPool<Worker, Data, Response> implements IPool

    interface IDynamicPool<Worker, Data, Response> extends AbstractPool

    class FixedThreadPool<Data, Response> extends AbstractPool
    class DynamicThreadPool<Data, Response> extends FixedThreadPool implements IDynamicPool

    class FixedClusterPool<Data, Response> extends AbstractPool
    class DynamicClusterPool<Data, Response> extends FixedClusterPool implements IDynamicPool
  @enduml
  ```
</details>



So to explain my pain-points with your current state of this PR:
At the start your PR should solve one problem: Providing a callback registration for the strategy of choosing a worker.

But now you rearranged the interfaces and abstractions and even changed the intention of the `IDynamicPool` interface always providing a `max` property

And if I see it correctly in your PR your solution also doesn't provide a better solution in terms of omit the use of binding functions and passing protected functions as callbacks to another function ([#191 L49](https://github.com/pioardi/poolifier/pull/191/files#diff-7d50e62835e81415b3afbe178f356673a4e28466fc12451ed99f1f33943d9455R49))

And (IMO) as a really bad side-effect you really increased the complexity of the UML of this project,
and in addition to that I think you produced a virtual bug if I'm not wrong.
Your `abstract class AbstractPool` implements `IPoolInternal` and your `interface IPoolInternal` implements `AbstractPool`. This is a circular implementation and I think other languages than TypeScript wouldn't even compile that.

I am programming since 2014 in business and have seen and worked with languages like C#, Java, Ruby and at least JS/TS.
TypeScript is my main-language since 2017, so I write and had build up knowledge about this language for more than 3 years now.

tl;dr: Specially the case that this PR just adds complexity and doesn't solve it's original purpose of just registering a choose worker strategy so no protected functions doesn't needed to be bounded anymore lets me decide to cannot agree with merging that.
If @pioardi is in another opinion and don't have problems with merging that, I just can say: He is the boss and author of this project.

---

_I'm sorry if some words are hard or written wrongly, but some of this could be due to I'm german native and therefore some english translations could be lost in my words._
_Also I don't want to attack you (@jerome-benoit) personally in any way, I enjoy working with you both together and we have done a great job around the last week(s)._

jerome-benoit:
> So to explain my pain-points with your current state of this PR:
> At the start your PR should solve one problem: Providing a callback registration for the strategy of choosing a worker.
>
> But now you rearranged the interfaces and abstractions and even changed the intention of the `IDynamicPool` interface always providing a `max` property

You can't provide such registration interface without making the same interface definition for pool. Try to do it from your initial PR and you will see by yourself. 

`IDynamicPool` is a pointless interface in any way. You shall not have to implement an interface for one pool type only. To properly implement design pattern strategy with callbacks registration, you require a unique interface to all pool types, it's a requirement. The way it's done in that PR is debatable but not its requirement: 

+ pool must have an interface fully describing the internal pool use and all pull implementations must implement it. All strategies (now and future) via callbacks require it to manipulate transparently all pool type classes (now and future). 
+ pool must have a public interface for end users that all pool types must implement. 
+ strategies callbacks consume the internal pool interface to manipulate pool classes.   
+ pool must not inherit one from another. 

And your classes diagram is not complete, all pools currently implements the IPool interface by inheritance. You can make the choice to state it explicitly in the code for readability by always adding the implements statement to classes. I do not care removing it but it makes the implementation of classes contracts implicit instead of explicit.   

> 
> And if I see it correctly in your PR your solution also doesn't provide a better solution in terms of omit the use of binding functions and passing protected functions as callbacks to another function ([#191 L49](https://github.com/pioardi/poolifier/pull/191/files#diff-7d50e62835e81415b3afbe178f356673a4e28466fc12451ed99f1f33943d9455R49))

Arrow function are bound to this by default. The use of the protected methods as arguments will go away once the term initial on that PR will not be anymore in its description ;)
Hint: the internal pool interface will allow the manipulate pool classes inside strategies callbacks transparently. 

> And (IMO) as a really bad side-effect you really increased the complexity of the UML of this project,
> and in addition to that I think you produced a virtual bug if I'm not wrong.
> Your `abstract class AbstractPool` implements `IPoolInternal` and your `interface IPoolInternal` implements `AbstractPool`. This is a circular implementation and I think other languages than TypeScript wouldn't even compile that.

It will compile with warnings that can silenced, interfaces are just attribute and method signatures that a classes must implement: IPoolInternal extends the **type** from the abstract, not the class. And it's common way to define interface from abstract class that need to be extended in all OO languages.  

I think you do not see yet the full picture of a proper strategy design pattern implementation via callbacks and the incremental path to reach it. The pool object manipulation inside external functions shall be avoided by sticking to the full OO implementation, but since strategy functions are currently factored out, let's reuse the factorisation first to allow to add strategies.
That PR is just a first step.

Shinigami92:
@jerome-benoit At first thanks not taking my comment as harassment or any kind of that :)

Give me some time over the weekend (maybe todays evening or tomorrow) so I can propose a draft on top of my PR parallel to your PR so we can compare each other
I will now enjoy a bit of free time in games and todays evening I'm in a weekly corona remote movie party with RL-friends ^^'

You could be right that I don't see the whole picture yet and maybe it is a new kind of design that I need to understand

---

Some additions:

- We don't need to rush with this
- That it only compiles with warnings that needs to be silenced hints me that we need to try out a better solution that doesn't compile with warnings
- I created both class diagrams with the exact written explicite extensions/implementations and implicit extensions/implementations are left out, that's correct
  But following the class diagrams it is totally okay to rely on implicit inheritance
  _Personally, I try to write as little code as possible, but to map all needed structural connections/inheritances correctly_

jerome-benoit:
> @jerome-benoit At first thanks not taking my comment as harassment or any kind of that :)

We are just having software engineering discussion, it's normal to have disagreements and to debate them to find out the best solution.

Shinigami92:
@jerome-benoit except that you are cheating :wink: and now we have the member functions `public`, I think this PR now moves in the very right direction :+1:  
As you are currently working on it I need to make a review when you are finished
Currently there are some warnings like missing JSDoc and missing _enforced_ return types etc

You can use `/** @inheritdoc */` for functions that already have JSDoc via their implemented interfaces
(https://github.com/gajus/eslint-plugin-jsdoc/issues/696#issuecomment-782012231)

You may also add `inheritdoc` to spellcheck list

@pioardi I think _sadly_ it is really the best way/option we have to just set member function access to `public`, because this simplifies so many things
Even if unknown contributors create their own pool implementations, they could develop their own strategies and provide them to us

jerome-benoit:
> @pioardi I think _sadly_ it is really the best way/option we have to just set member function access to `public`, because this simplifies so many things
> Even if unknown contributors create their own pool implementations, they could develop their own strategies and provide them to us

If you look at the transpiled files, the internal interfaces are not directly exported, the abstract class for pool neither. So for TS use, that does not change anything unless you import some modules with the relative path directly to the file. But you can always bypass the module export and classes components scope that way by using JS. 

One way to avoid changing the scope on methods on pool, is to only pass the workers array and the tasks map, and return only the index of the chosen worker or null to init a worker creation and the potential worker to destroy. That would the API really cleaner. But I do not plan to make it in that PR.

Shinigami92:
It is `public`

image

PR #201, opened 2021-02-19
JSONValue can not be used by custom defined interfaces

jerome-benoit:

## PR Checklist

- [ ] Yes
- [x] No

---


Shinigami92 (changes requested):
Why does VSCode present me an error for this?

image

Also as can be seen in the tooltip, the type falls back to `any` :warning:

jerome-benoit:
> Why does VSCode present me an error for this?
> 
> image
> 
> Also as can be seen in the tooltip, the type falls back to `any` ⚠️

I do not know but it's a bug in vscode, recursive type definition are supported and working in TS.

Shinigami92:
Then we should find the bug issue in VSCode or if there is no issue we should open one
Currently it's really not helpful to see a type that is just `any`

Shinigami92:
@jerome-benoit I don't think you are right here...

Even TypeScript Playground shows me this error!

https://www.typescriptlang.org/play?ts=4.2.0-beta#code/C4TwDgpgBAaghgGwK4QPICcCC71xAHgBUA+KAXikKgB8ptcD5k0sc8jjiBuAKB9EhQAUgGVUAOSYpyPKDSgBnYOgCWAOwDms+WqQBbAEYR022gYD25hBDhrTUXQgT3REqRADaAXXsAlCADG5ugAJvhKqpoANMJikogoxEA

The change in TS 3.7 https://devblogs.microsoft.com/typescript/announcing-typescript-3-7/#more-recursive-type-aliases seems to have only fixed generic types recursion and NOT direct circular references

The type is therefor building, but an error and is falling back to just `any`

If you really want that to work, you should create an issue to Microsoft in TypeScript repo

I don't want have this merged before it is correctly usable in an IDE like VSCode

jerome-benoit:
> @jerome-benoit I don't think you are right here...
> 
> Even TypeScript Playground shows me this error!
> 
> [typescriptlang.org/play?ts=4.2.0-beta#code/C4TwDgpgBAaghgGwK4QPICcCC71xAHgBUA+KAXikKgB8ptcD5k0sc8jjiBuAKB9EhQAUgGVUAOSYpyPKDSgBnYOgCWAOwDms+WqQBbAEYR022gYD25hBDhrTUXQgT3REqRADaAXXsAlCADG5ugAJvhKqpoANMJikogoxEA](https://www.typescriptlang.org/play?ts=4.2.0-beta#code/C4TwDgpgBAaghgGwK4QPICcCC71xAHgBUA+KAXikKgB8ptcD5k0sc8jjiBuAKB9EhQAUgGVUAOSYpyPKDSgBnYOgCWAOwDms+WqQBbAEYR022gYD25hBDhrTUXQgT3REqRADaAXXsAlCADG5ugAJvhKqpoANMJikogoxEA)

Then TS playground have the same bug. It probably uses the same code analyser than vscode.

```ts
type Record<K extends keyof any, T> = {
    [P in K]: T;
};
```

And

```ts
type JSONValue =
  | string
  | number
  | boolean
  | null
  | JSONValue[]
  | [k: string]: JSONValue
```

does not have a circular reference error but

```ts
type JSONValue =
  | string
  | number
  | boolean
  | null
  | JSONValue[]
  | Record<string, JSONValue>
```
 
does have wrongly one. 

That make no sense at all and the TS transpiler fails in case of circular reference. That's not the case here. 

> The change in TS 3.7 [devblogs.microsoft.com/typescript/announcing-typescript-3-7/#more-recursive-type-aliases](https://devblogs.microsoft.com/typescript/announcing-typescript-3-7/#more-recursive-type-aliases) seems to have only fixed generic types recursion and NOT direct circular references

The type definition here is a recursive one. The circular reference have a complete other meaning and is not present in the that type definition.

pioardi:
Let me say that I really don't know typescript, but even nodejs doc contains any and the problem that jerome is having is the main reason why I would prefer the type any :) We can then specify into our doc that only some kind of data should be passed as per nodejs doc

src/utility-types.ts
```diff
@@ -7,23 +7,16 @@ import type { KillBehavior } from './worker/worker-options'
...
- * Serializable JSON array.
- */
-export type JSONArray = Array<JSONValue>
+export type JSONValue =
+  | string
+  | number
+  | boolean
+  | null
+  | JSONValue[]
+  | Record<string, JSONValue>
```
Shinigami92, line 19: So this fix should work for now, until Microsoft has fixed it

```suggestion
  | { [k: string]: JSONValue }
```

Shinigami92:
> Let me say that I really don't know typescript, but even nodejs doc contains any and the problem that jerome is having is the main reason why I would prefer the type any :) We can then specify into our doc that only some kind of data should be passed as per nodejs doc

So you mean we should get rid of the json definition completely and let the customer of your plugin fail on their own?
Please make it really clear in e.g. the JSDoc that only serializable data can be passed
This cost us half a day with theoretically 3 developers back in the days because we didn't have this knowledge and didn't thought about it

Shinigami92:
All I want is to change `Record<string, JSONValue>` to `{ [k: string]: JSONValue }`
It has the exactly same meaning but works in IDEs
`Record` is just a wrapper util type that does exactly this what `{ [k: string]: something }` is

pioardi:
Since I don't know TS really well I will leave this to you....

src/utility-types.ts
jerome-benoit: With that type definition, you get a missing Index compilation error with a 
WorkerData custom interface like `{ index: number, data: any }` which is perfectly legal. 
With the Record type, it just works as expected. 
I do see the point of not using the correct, bug free, upstream maintained and working type definition shortcuts because of bugs that does not belong to poolifier. 
The code change is that PR made poolifier 2.0.0 just usable, without it's not usable.

Shinigami92:
A better example https://www.typescriptlang.org/play?#code/PTAEHEFMDtIJwJYGNRIXJBXANgQzqAgM6gDuA9nANYLQDmAUAC4CeADpKAGq7aaQB5OAEE4cXCwA8AFQB8oALyhpoAD6hR4qTz6CRYiTNmyA3AwYhQAOXKgAJukhImoVh1ToseRK2btOAFIAygJWOvyKDKBqoERMiPRRMdCYALYARvBJ6unk5NiQuNDZoCnY2CXBoeGQANoAuiUA3qC1VABcsfG0dPWdVWG8EQC+5pYAqkQ9oOSk0KAASk6UdvaQAGa0CEwI5POk2wAWXQl0JLgkVJAsRH7uQd30S0grRoqgLW2dcad9yqCjNyBEKDXQAJki0XUPx6JRSGSyUNAuXyhWKSLKFSRAxqYIaJQep2erxxQ0gYNkYzAk2mxLgdjuwOqZIAzJCYjDEhi0pk4CUUQUinCcFiYqTdCz8Ui6XZJJy6AAaUDi-gsylAA

src/utility-types.ts
Shinigami92: THAT one is the real bug in our tool :eyes: 
damnd!

Okay so @pioardi are unintentionally right and we need to use unknwon everywhere + use jsdoc

Shinigami92: @jerome-benoit I will make the change in this branch, so please give me a minute or two

jerome-benoit: Well, I prefer to have type for workerData that works than any.

Shinigami92: Defining `WorkerData` on your own is exactly possible and then should be used as the first generic parameter for the Worker and the Pool

Shinigami92: So like this: 
image

jerome-benoit: Fair enough. Let the user enforce its in/out data types.

Shinigami92 (dismissed):

pioardi (dismissed):
That's better for me now :P

Shinigami92:
@pioardi So jerome found out that the JSONValue definition cant be used at all by customers :warning: 
I removed it completely and used `unknown` as default

I also added JSDoc to every data usage so the developer hopefully see it
We may add a hint also to the readme :thinking:

jerome-benoit:
I cannot approve my own PR but @Shinigami92 changes are fine with me :)

Shinigami92 (approved):

Shinigami92:
Added a hint to the readme

@pioardi if you approve it, you can also merge it now :+1:

README.md
```diff
@@ -137,6 +137,8 @@ You can do the same with the classes ClusterWorker, FixedClusterPool and Dynamic
 **See examples folder for more details (in particular if you want to use a pool for [multiple functions](./examples/multiFunctionExample.js)).**  
 **Now TypeScript is also supported, find how to use it into the example folder**.
 
+Remember that workers can only send and receive serializable (JSON) data.
```
pioardi, line 140: I am not sure that this is totally accurate based on node js doc, I need more time to look into that.

Shinigami92: At least it's a placeholder / reminder for us to improve the readme on this point ^^
jerome already moved forward :shrug: 
But I don't think that this is a problem

pioardi: No problem we can always be more detailed if we figure out that is needed :) it is ok

Shinigami92:
You want to let this here for clearance 

image

PR #210, opened 2021-02-20
Add a less used worker choice strategy

jerome-benoit:
And cleanup the implementation of worker selection strategies.

Signed-off-by: Jérôme Benoit <jerome.benoit@sap.com>

## PR Checklist

- [x] Please add a description of your changes.
- [ ] We need your changes to come with unit tests in order to keep this project in quality and easy to maintain.
- [ ] Please add a link to the open issue or task that this pull request will resolve.
- [ ] Add a note to the changelog to indicate the change.

## Does this PR introduce a breaking change?

- [ ] Yes
- [x] No

If yes please indicate it and migration info into the CHANGELOG.md file.

---

PR #213, opened 2021-02-21
Check that a pool have a minimum number of workers

jerome-benoit:

## PR Checklist

- [ ] Yes
- [x] No

---

https://github.com/pioardi/poolifier/pull/210#discussion_r579787739


Shinigami92:
What's with **dynamic** pools where `numberOfWorkers === 0`?

Shinigami92:
Please also add tests for at least following cases:
1. Verify the check itself
2. Verify that a dynamic pool can have min 0 workers
3. Verify that a fixed pool must have at least 1 worker

jerome-benoit:
> Please also add tests for at least following cases:
> 
> 1. Verify the check itself
> 2. Verify that a dynamic pool can have min 0 workers
> 3. Verify that a fixed pool must have at least 1 worker

We first need to merge the PR that implement strategies on worker choice and add `isDynamic()` method. Otherwise code will be duplicated.

pioardi (dismissed):

Shinigami92:
@pioardi You don't need to approve draft PRs because your approval gets stale when new commits come into it

jerome-benoit:
I do not understand why sonarcloud coverage is failing, the branches added are checked in unit tests.

src/pools/abstract-pool.ts
```diff
@@ -166,6 +167,24 @@ export abstract class AbstractPool<
     }
   }
 
+  private checkNumberOfWorkers (numberOfWorkers: number): void {
+    if (numberOfWorkers === undefined) {
```
Shinigami92, line 171:
```suggestion
    if (numberOfWorkers == null) {
```

src/pools/abstract-pool.ts
```diff
@@ -166,6 +167,24 @@ export abstract class AbstractPool<
...
+      )
+    } else if (!Number.isSafeInteger(numberOfWorkers)) {
+      throw new Error(
+        'Cannot instantiate a pool with a non integer number of workers'
+      )
+    } else if (numberOfWorkers < 0) {
+      throw new Error(
+        'Cannot instantiate a pool with a negative number of workers'
+      )
+    } else if (!this.isDynamic() && numberOfWorkers <= 0) {
```
Shinigami92, line 183:
```suggestion
    } else if (!this.isDynamic() && numberOfWorkers === 0) {
```

or

```suggestion
    } else if (!this.isDynamic() && numberOfWorkers < 1) {
```

tests/pools/abstract/abstract-pool.test.js
```diff
@@ -1,5 +1,6 @@
 const expect = require('expect')
 const { FixedThreadPool } = require('../../../lib/index')
+const { FixedClusterPool } = require('../../../lib/index')
```
Shinigami92, line 3: I think you can merge this to `const { FixedClusterPool, FixedThreadPool } = require('../../../lib/index')`

Shinigami92:
@jerome-benoit Did you know you can apply the suggested changes via the GitHub Web interface? On the files tab you can also apply the changes in batch at once.

---

Theoretically the code is okay
But we need to find out why the coverage dropped to 68.4%
This way it is not acceptable to merge currently :slightly_frowning_face:

jerome-benoit:
It's a bug on the static analyser.

tests/pools/abstract/abstract-pool.test.js
```diff
@@ -61,4 +61,40 @@ describe('Abstract pool test suite', () => {
       const pool = new StubPoolWithIsMainMethod(1, '').toThrowError()
```
Shinigami92, line 61: This test doesn't look right to me :thinking:  
the `.toThrowError()` should be called on `expected` and not inside it

I could be wrong, but also I think just calling `.toThrowError()` with no `new Error(...)` argument always returns `true` :confused:  
This is really odd by mocha but I think it is this way :slightly_frowning_face:

tests/pools/abstract/abstract-pool.test.js
```diff
@@ -61,4 +61,40 @@ describe('Abstract pool test suite', () => {
...
     })
   })
+
+  it('Verify that numberOfWorkers is checked', () => {
+    expect(() => {
+      const pool = new FixedThreadPool(
+        './tests/worker-files/thread/testWorker.js'
+      ).toThrowError(
```
Shinigami92, line 69: And this may be the error here :warning: 
It should be called on expected and not inside it! The tests are running without doing anything!

Shinigami92 (changes requested):

Shinigami92:
@jerome-benoit You need to add a new test-case for `"Cannot instantiate a pool without specifying the number of workers"`

tests/pools/abstract/abstract-pool.test.js
```diff
@@ -42,59 +42,54 @@ describe('Abstract pool test suite', () => {
...
+    )
   })
 
   it('Verify that numberOfWorkers is checked', () => {
-    expect(() => {
-      const pool = new FixedThreadPool(
-        './tests/worker-files/thread/testWorker.js'
-      ).toThrowError(
-        new Error(
-          'Cannot instantiate a pool without specifying the number of workers'
```
Shinigami92: The old test-case for this did never run and it was not correct

Shinigami92:
Now that we already check src directly instead of lib, should we write our tests also in typescript? That way the bug would be seen much better and earlier because the function toThrowError() does not exist on the pool but the return of expect

Shinigami92:
Ah so it was a copy paste error beforehand I think

Shinigami92:
`100.0% Coverage` great job to us :tada:

pioardi:
It looks great, I will need more time to review that better.
Feel free to go ahead if both of you are confident :)

Shinigami92 (approved):

Shinigami92:
> It looks great, I will need more time to review that better.
> Feel free to go ahead if both of you are confident :)

We are not in a haste, so if you want a second approval on your side, we should do that

jerome-benoit:
> It looks great, I will need more time to review that better.
> Feel free to go ahead if both of you are confident :)

It's just sanity checks for pool creation parameters. 

I think the next step can be to provide default values if they are not provided: 
+ fixed: numberOfThreads = number of CPUs the OS see
+ dynamic: min = 0 and max = number of CPUs the OS see

through a kind of configurations mask object for pool parameters.

Shinigami92 (approved):

Shinigami92:
> I think the next step can be to provide default values if they are not provided:
> 
> * fixed: numberOfThreads = number of CPUs the OS see
> * dynamic: min = 0 and max = number of CPUs the OS see
> 
> through a kind of configurations mask object for pool parameters.

I'm not sure if this would be good :thinking:  
A ThreadPool could run more threads that CPUs in OS
and maybe a ClusterPool should not use the whole system by default
If users want to do this, they should just pass the values and read them out on their own
We could provide examples for that and document it, but default values could not be so good

pioardi (approved):
It is a great job, thank you guys for that

pioardi:
> > It looks great, I will need more time to review that better.
> > Feel free to go ahead if both of you are confident :)
> 
> It's just sanity checks for pool creation parameters.
> 
> I think the next step can be to provide default values if they are not provided:
> 
> * fixed: numberOfThreads = number of CPUs the OS see
> * dynamic: min = 0 and max = number of CPUs the OS see
> 
> through a kind of configurations mask object for pool parameters.

We can think about that later on yes :)

PR #222, opened 2021-02-22
Add a random worker selection once the pool is full strategy.

jerome-benoit:
Not sure it will be useful for a real world workload but it was quick to
implement ;)

## PR Checklist

- [ ] Yes
- [x] No

---


pioardi:
Thanks for your submission Jerome, I Will look more deeply tomorrow but please note that there are two security hotspots on sonar

src/pools/selection-strategies.ts
```diff
@@ -106,25 +110,41 @@ class LessRecentlyUsedWorkerChoiceStrategy<
...
+        this.pool
       )
+      if (freeWorker) {
+        return freeWorker
+      }
+    }
+    return this.pool.workers[this.getRandomInt(0, this.pool.workers.length - 1)]
+  }
+
+  private getRandomInt (max: number, min = 0): number {
```
Shinigami92, line 143: Should we use this implementation?
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/random#getting_a_random_integer_between_two_values

src/pools/selection-strategies.ts
```diff
@@ -94,7 +98,7 @@ class LessRecentlyUsedWorkerChoiceStrategy<
     // A worker is always found because it picks the one with fewer tasks
     let lessRecentlyUsedWorker!: Worker
     for (const [worker, numberOfTasks] of this.pool.tasks) {
-      if (numberOfTasks === 0) {
+      if (!this.pool.isDynamic() && numberOfTasks === 0) {
```
Shinigami92, line 101: We may extract this `!this.pool.isDynamic()` out of the loop to a `const` and safe performance this way, cause the function is not called for each task

Shinigami92 (changes requested):

pioardi:
I am not sure we really Need this guys, i am thinking to a real scenario where this can be really a benefit mmmmm

Shinigami92:
> Thanks for your submission Jerome, I Will look more deeply tomorrow but please note that there are two security hotspots on sonar

It's because we call `Math.random()`. But we are not affected by any vulnerabilities and have not security code here. So you @pioardi can go to SonarCloud and just disable this warning.

Shinigami92:
> I am not sure we really Need this guys, i am thinking to a real scenario where this can be really a benefit mmmmm

I also thought that :thinking:  
The code itself is okay, but the need... :shrug: Not really :thinking:

src/pools/selection-strategies.ts
jerome-benoit: The implementation in the PR is more optimized: one floor, one random only.

jerome-benoit:
> I am not sure we really Need this guys, i am thinking to a real scenario where this can be really a benefit mmmmm

It's useful for testing purpose, especially for internal benchmarking on selection strategies as a kind of worst case.

src/pools/selection-strategies.ts
Shinigami92: But the need of an if else branch :thinking:  
An internal benchmark could be nice here again

Shinigami92: Ah you removed the branch

pioardi:
> > I am not sure we really Need this guys, i am thinking to a real scenario where this can be really a benefit mmmmm
> 
> It's useful for testing purpose, especially for internal benchmarking on selection strategies as a kind of worst case.

mmmmm I really don't see this needed for testing purposes, I mean why we need a worst case ?
I am sorry I just really do not see any benefit providing this feature, rather it could be misleading for end users

Shinigami92:
I also agree with @pioardi 
Also, and now you see hopefully the problem, you are doing multiple things in one PR again.
Some little code snippets are okay but they are in this big branch and you need to extract them now into a new PR if we want them :slightly_frowning_face:  

@jerome-benoit If you provide a new feature (like a new strategy) you should also think about creating an issue for that so we can talk about it before wasting time implementing it
Instead we could implement a way that users can pass their own strategy. Currently the switch-case only allows poolifier internal strategies and prevent the user from trying they own strategy

jerome-benoit:
> > > I am not sure we really Need this guys, i am thinking to a real scenario where this can be really a benefit mmmmm
> > 
> > 
> > It's useful for testing purpose, especially for internal benchmarking on selection strategies as a kind of worst case.
> 
> mmmmm I really don't see this needed for testing purposes, I mean why we need a worst case ?
> I am sorry I just really do not see any benefit providing this feature, rather it could be misleading for end users

It's a performance evaluation thing, it's like fault injection for security: you do things unexpected in a code path and you fix its impact. And you keep it as a landmark for non regression on the code path.   

It's ok to change the API to allow to pass that strategy for performance evaluation instead of providing it in poolifier.

jerome-benoit:
> @jerome-benoit If you provide a new feature (like a new strategy) you should also think about creating an issue for that so we can talk about it before wasting time implementing it
> Instead we could implement a way that users can pass their own strategy. Currently the switch-case only allows poolifier internal strategies and prevent the user from trying they own strategy

It's way more efficient to discuss around a piece of code than around an issue. 
GitHub implied code flow is not efficient but I can't change it. I will try to use stgit in a future to split implementation in small logical pieces that could be submitted as a GitHub PR (the inefficient GitHub understanding of it ... unfortunately)

Shinigami92:
@jerome-benoit I think you are confusing `efficient` with `fast`

pioardi:
> > > > I am not sure we really Need this guys, i am thinking to a real scenario where this can be really a benefit mmmmm
> > > 
> > > 
> > > It's useful for testing purpose, especially for internal benchmarking on selection strategies as a kind of worst case.
> > 
> > 
> > mmmmm I really don't see this needed for testing purposes, I mean why we need a worst case ?
> > I am sorry I just really do not see any benefit providing this feature, rather it could be misleading for end users
> 
> It's a performance evaluation thing, it's like fault injection for security: you do things unexpected in a code path and you fix its impact. And you keep it as a landmark for non regression on the code path.
> 
> It's ok to change the API to allow to pass that strategy for performance evaluation instead of providing it in poolifier.

I think having the possibility to pass a strategy and then pass the random strategy from a test would be a great trade-off.

jerome-benoit:
> @jerome-benoit I think you are confusing `efficient` with `fast`

Not at all, have a look at the way PRs are done on git, Linux Kernel, OpenWrt (when GitHub is not used), glibc, etc. PRs here huge and review is done at the commit level.

Shinigami92:
> > @jerome-benoit I think you are confusing `efficient` with `fast`
> 
> Not at all, have a look at the way PRs are done on git, Linux Kernel, OpenWrt (when GitHub is not used), glibc, etc. PRs here huge and review is done at the commit level.

You compare big-sizes repos with great amount of developers with a 3-member small-sized JS-lib :eyes:

jerome-benoit:
> > > @jerome-benoit I think you are confusing `efficient` with `fast`
> > 
> > 
> > Not at all, have a look at the way PRs are done on git, Linux Kernel, OpenWrt (when GitHub is not used), glibc, etc. PRs here huge and review is done at the commit level.
> 
> You compare big-sizes repos with great amount of developers with a 3-member small-sized JS-lib 👀

A code flow does not change because of the community size or code size, it changes because of the tooling and project rules regarding it. 
When you use GitHub, you are stuck to PR based level review and that's not efficient.

pioardi:
@jerome-benoit  if you can jump into the chat I would love a discussion about the last message you sent :)

Shinigami92:
IMO this PR is superseded by #234  
I will close it

PR #226, opened 2021-02-23
Coverage increased

pioardi:
## PR Checklist

- [x] We need your changes to come with unit tests in order to keep this project in quality and easy to maintain.

- [ ] Yes
- [x] No

---

Increase coverage adding more unit tests


pioardi:
@jerome-benoit @Shinigami92 I would let you add unit tests for strategies if you have time today into this PR or another.
After this other unit tests we only need to document selection strategies and we can then release the stable 2.0.0 version

Shinigami92:
> @jerome-benoit @Shinigami92 I would let you add unit tests for strategies if you have time today into this PR or another.
> After this other unit tests we only need to document selection strategies and we can then release the stable 2.0.0 version

I don't know if I will have time before ~15:00
And even then I will maybe rest a bit after work in some games
So currently maybe only in review-mode

pioardi:
Nevermind I will let you go though that when you feel is needed.
The only point where the coverage is missing on the selection strategies are the default values.

tests/worker/abstract-worker.test.js
```diff
@@ -1,10 +1,28 @@
...
+
+  it('Verify that handle Error function is working properly', () => {
+    const error = new Error('My error')
+    const worker = new ThreadWorker(() => {})
+    expect(worker.handleError(error)).toBe(error)
+  })
+
+  it('Verify that get main worker throw error if main worker is not set', () => {
+    const worker = new StubPoolWithIsMainWorker(() => {})
+    expect(() => worker.getMainWorker()).toThrowError()
```
jerome-benoit, line 26: You can remove the useless intermediate worker var: 

expect(() => new StubPoolWithIsMainWorker(() => {}).getMainWorker()).toThrowError()

jerome-benoit (changes requested):

tests/worker/abstract-worker.test.js
pioardi: I see you are starting to appreciate sonar :P

pioardi: Done

jerome-benoit (approved):

PR #243, opened 2021-02-27
Return false instead of null tuple

Shinigami92:
Little improvement to prevent the usage of `JSON.stringify`

### DIFF
--- a/src/pools/selection-strategies.ts
+++ b/src/pools/selection-strategies.ts
@@ -137,9 +137,9 @@ class DynamicPoolWorkerChoiceStrategy<Worker extends IWorker, Data, Response>
 
   /** @inheritdoc */
   public choose (): Worker {
-    const freeWorkerEntry = this.pool.findFreeTasksMapEntry()
-    if (freeWorkerEntry) {
-      return freeWorkerEntry[0]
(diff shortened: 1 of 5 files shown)


src/pools/selection-strategies.ts
```diff
@@ -137,9 +137,9 @@ class DynamicPoolWorkerChoiceStrategy<Worker extends IWorker, Data, Response>
 
   /** @inheritdoc */
   public choose (): Worker {
-    const freeWorker = this.pool.findFreeTasksMapEntry()[0]
-    if (freeWorker) {
-      return freeWorker
+    const freeWorkerEntry = this.pool.findFreeTasksMapEntry()
+    if (freeWorkerEntry) {
+      return freeWorkerEntry[0]
```
jerome-benoit, line 142: It's a freeTaskMapEntry. TS transpiler is Ok? Because I have had an error here while testing the same.

Shinigami92: Check it out and test it on your local setup
Maybe you just missed something when trying yourself, but now everything is just working (on my side)

jerome-benoit (approved):

PR #364, opened 2021-05-16
Dedupe worker attributes

jerome-benoit:

## PR Checklist

- [ ] If your changes contain any new feature please be sure to document it.
- [x] Add a note to the changelog to indicate the change.

- [x] Yes
- [ ] No

---

### DIFF
--- a/tests/worker/abstract-worker.test.js
+++ b/tests/worker/abstract-worker.test.js
@@ -15,11 +15,11 @@ describe('Abstract worker test suite', () => {
     )
   })
 
-  it('Verify worker default values', () => {
+  it('Verify worker options default values', () => {
     const worker = new ThreadWorker(() => {})
-    expect(worker.maxInactiveTime).toBe(1000 * 60)
-    expect(worker.killBehavior).toBe(KillBehaviors.SOFT)
-    expect(worker.async).toBe(false)
+    expect(worker.opts.maxInactiveTime).toBe(1000 * 60)
+    expect(worker.opts.killBehavior).toBe(KillBehaviors.SOFT)
+    expect(worker.opts.async).toBe(false)
   })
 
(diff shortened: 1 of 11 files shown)


tests/worker/abstract-worker.test.js
```diff
@@ -15,11 +15,11 @@ describe('Abstract worker test suite', () => {
...
 
-  it('Verify worker default values', () => {
+  it('Verify worker options default values', () => {
     const worker = new ThreadWorker(() => {})
-    expect(worker.maxInactiveTime).toBe(1000 * 60)
-    expect(worker.killBehavior).toBe(KillBehaviors.SOFT)
-    expect(worker.async).toBe(false)
+    expect(worker.opts.maxInactiveTime).toBe(1000 * 60)
+    expect(worker.opts.killBehavior).toBe(KillBehaviors.SOFT)
+    expect(worker.opts.async).toBe(false)
```
Shinigami92, line 22: **question:** So this is a breaking change to the public / community that uses our package, isn't it? Is it necessary to introduce this change?

jerome-benoit: The attributes are not public. 
If there's users relying on it, it's their responsibility to not use protected attributes of an abstract class and not yours.
I can document the change but nobody needs to use that abstract class.

Shinigami92: Oh so they were accessed here, cause JS didn't stopped us...
We should write tests also in TypeScript :/

Shinigami92 (dismissed):

tests/worker/abstract-worker.test.js
jerome-benoit: Nevertheless, I will document the change for a 2.1.x release, the AbstractWorker class is exported.

Shinigami92 (approved):

PR #439, opened 2021-07-10
Apply dependencies update

jerome-benoit:

## PR Checklist

- [ ] If your changes contain any new feature please be sure to document it.

- [ ] Yes
- [x] No

---


src/pools/abstract-pool.ts
```diff
@@ -100,7 +100,8 @@ export abstract class AbstractPool<
   Worker extends IWorker,
   Data = unknown,
   Response = unknown
-> implements IPoolInternal<Worker, Data, Response> {
+> implements IPoolInternal<Worker, Data, Response>
```
Shinigami92, line 103: I'm unsure if you forget to run prettier after its update or eslint is configured wrongly :thinking:  
This or that, what do you think about `git hooks`?

jerome-benoit: It's the result of `npm run format`.

Shinigami92: Okay, I don't care anymore :shrug: 
IMO the project setup for prettier is already to complex
And I'm not so much interested in poolifier anymore

IMO @pioardi Should just give you owner rights and you can potentially overtake this repo

Shinigami92 (approved):

src/pools/abstract-pool.ts
jerome-benoit: Something have probably changed in the latest prettier or prettierx version. I think eslint still expect the parenthesis to be on the same line, but I guess eslint can be configured to expect the same. 

I hope your interest loss in poolifier is temporary ... 

If you find an interest to fix it, you can make a PR on top of that one ;)

Shinigami92: I think eslint may just respect prettier default but not prettierx and these "standard" formatting rules that really few projects depend on out there
prettier is intended to format code in a way so teams don't need to talk about formatting anymore, but manipulating it with stuff like prettierx and forcing it to do what you (@pioardi) personally wants breaks things like this

So maybe we could remove prettier and let chaos just being (and so I will also leave the project), or we could try to configure it that it just works out of the box without extra stuff like _"standard"_
But even in the second path, I personally current have any projects anymore that uses poolifier and I also don't have the knowledge about clusters and pools and stuff and I'm no node developer but a frontend developer writing single page applications...

And I'm involved in many other projects like vite, @prettier/plugin-pug, eslint-define-config and other stuff
So yeah, I can try to review poolifier from time to time, but I don't know if I could maintain it, you know...

Shinigami92: My frustration doesn't only come from prettier(x) and so on, but also the not working pipeline/ci setup
We could try renovate instead, but I think this will result in the same problem from sonarcloud

And pioardi is so much absend that I'm question myself if he wants to give the project into your hands
You jerome do such a great job and keep the repo up and running

jerome-benoit: For the formatting, I think the project can live with relying on `npm run format` as the correct formatting for the time being. And fix eslint later on. 

For the other topics, I can't say more about it. I think an organisation should be created to allow real co-maintenance with @pioardi (rights issue with repo under personal account
) and that repo to be migrated under it. And access to CI tools given also to me so I will be able to fully co-maintain poolifier and fix issues when @pioardi is not available.

